PR #2, opened 2025-07-07
TRA458: Sleep-In Logic – Implement sleepIn(weekday, vacation)

haithemhamid:
The sleepIn method in the TRA458 class checks whether a person can sleep in based on two conditions: if it's not a weekday or if the person is on vacation, then they can sleep in (returns true); otherwise, they can't (returns false).

### DIFF
--- a/plag_test_tool/from_Haitham/TRA458.java
+++ b/plag_test_tool/from_Haitham/TRA458.java
@@ -1,0 +1,10 @@
+public class TRA458{  
+public static boolean sleepIn(boolean weekday, boolean vacation) {
+        if (!weekday) {
+            return true;
+        } else if (vacation) {
+            return true;
+        }
+        return false;
+    }
+	}


EchoCoder20 (changes requested):
Line 1: Class names should be descriptive and match the file name. The class name 'SquirrelActivity' is appropriate here because it describes the activity related to the squirrels. Make sure the class name and file name remain consistent.
Line 2: The function name 'isSquirrelPlaying' effectively communicates its purpose, which is to determine if squirrels are playing based on the temperature and the season.
Line 2: Parameter names 'temp' and 'isSummer' are descriptive enough, indicating temperature and whether it is summer. Ensure variable names are always clear in their purpose.
Line 3: Ensure the code is properly formatted with consistent indentation. Remove any unnecessary blank lines to maintain clarity and conciseness.
Line 13: Add a newline at the end of the file to adhere to the best practices and conventions in Java.

EchoCoder20 (changes requested):
{
  "body": ,
  "event": 
}

PR #4, opened 2025-07-07
#tra 461:Sum Double – Return Sum or Double the Sum

sh3tha:
Sum Double – Return Sum or Double the Sum


daniyalmehmood:
@sh3tha Please follow proper naming conventions as discussed

haithemhamid (approved):
make sure to add each task separate and same things about making proper descripiton.

PR #7, opened 2025-07-07
#tra 462:Diff 21 – Return Absolute Difference or Double If Over 21

sh3tha:
Diff 21 – Return Absolute Difference or Double If Over 21


haithemhamid (approved):
good

PR #9, opened 2025-07-07
TRA491: Sum Double – Return Sum or Double the Sum

haithemhamid:
The sumDouble method adds two numbers a and b, but if the numbers are the same, it returns double their sum; otherwise, it just returns the regular sum.

### DIFF
--- a/plag_test_tool/from_Haitham/TRA491.java
+++ b/plag_test_tool/from_Haitham/TRA491.java
@@ -1,0 +1,11 @@
+public class TRA491{
+public static int sumDouble(int a, int b) {
+        int sum = 0;
+        if (a == b) {
+            sum = 2 * (a + b);
+
+        } else if (a != b) {
+            sum = a + b;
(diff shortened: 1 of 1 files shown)


haithemhamid:
> Good, But few suggestion to improve your code:
> 
> The condition else if (a !=b) is not necessary because if (a==b) is false so a!=b will be by default true so, you can use else instead of else if

I put else if to know in future why this condition I have it.

EchoCoder20:
> > Good, But few suggestion to improve your code:
> > The condition else if (a !=b) is not necessary because if (a==b) is false so a!=b will be by default true so, you can use else instead of else if
> 
> I put else if to know in future why this condition I have it.

You can use comment to know why you have this condition. Also, using else instead of else if  will simplify the logic without changing the result.

NASSER1221 (approved):
*you can use else directly to simplify.
*int sum is not needed.

PR #10, opened 2025-07-07
 TRA-601 : Solving TRA-601 task

EchoCoder20:
Implement a method diff21(int n) that returns the absolute difference between the given number n and 21.
If the number is greater than 21, return double the absolute difference.


Ghaida02a:
sync with master

Ghaida02a (approved):

PR #14, opened 2025-07-08
TRA-602: Solving Parrot Trouble task

EchoCoder20:
I implement method that check for talking parrot at bad hours. I passed two parameters to the method (talking and hours).
If parrot talking and hours more than 20 or less than 7 then I return true otherwise false is return


haithemhamid (dismissed):

Ghaida02a:
sync with master

Ghaida02a (approved):

PR #16, opened 2025-07-08
TRA-465: Solving Makes 10 task

EchoCoder20:
I Implement a method called  makes10 that passed two parameters a and b that returns true if:

- Either a is 10 or b is 10, or

- The sum of a and b is 10

otherwise it return false


haithemhamid (approved):

Ghaida02a (changes requested):
Well done

PR #22, opened 2025-07-08
Sleep-In Logic – Implement sleepIn(weekday, vacation)

NASSER1221:
*Used if statement 
*used Boolean type to determine the status if it is a vacation or weekday.


EchoCoder20 (changes requested):
Your code logic is correct , but instead of write the expression inside if statment like this : `if(aSmile==True)` you need to write if like this `if(isSmile)` beacuse `if(aSmile==true)` is equle to `if(aSmile) `and `if(aSmile==false)` is same as `if(!aSmile)`

NASSER1221:
changed according to the request

PR #24, opened 2025-07-08
TRA-470:Solving missing char task

EchoCoder20:
I Implement a method called missingChar that pass string and integer param and returns a new string where the character at index integer param has been removed.

### DIFF
--- a/Horiya_Albusaidi/MissingChar.java
+++ b/Horiya_Albusaidi/MissingChar.java
@@ -1,0 +1,14 @@
+public class MissingChar {
+    public static String missingChar(String str, int n)
+    {
+        StringBuilder modifiedString=new StringBuilder(str);
+
+        if(n>=0 && n<str.length())
+        {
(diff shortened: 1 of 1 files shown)


haithemhamid (approved):

From_Horiya_Albusaidi/MissingChar.java
```diff
@@ -0,0 +1,16 @@
...
+        StringBuilder modifiedString=new StringBuilder(str);
+        if (str == null) {
+            return "The input string cannot be null";
+        }
+        if(n>=0 && n<str.length())
+        {
+            modifiedString.deleteCharAt(n);
+            return modifiedString.toString();
+        }
+        else return "Invalid index: must be between 0 and " + (str.length() - 1)
```
Ghaida02a, line 13: else return "Invalid index: must be between 0 and " + (str.length() - 1);
should add (;)

Ghaida02a (changes requested):

PR #25, opened 2025-07-08
Tra 463: Added Sleep-In Logic Java Code for Ghaida 

Ghaida02a:
Implement SLeepIn(boolean weekday, boolean vacation) Function to check if it not a weekday or vacation, it will return true (we can sleep). Otherwise, false(we cannot sleep)


EchoCoder20 (approved):
Great work

haithemhamid (approved):

PR #26, opened 2025-07-08
TRA459: Monkey Trouble Coding 

hanin190340:
Implement a method monkeyTrouble(boolean aSmile, boolean bSmile) that determines if we are in trouble based on whether two monkeys are smiling. We are considered in trouble if both monkeys are smiling or if neither of them is smiling; otherwise, we are not in trouble. The goal is to practice using boolean logic and logical operators to return true when both aSmile and bSmile are the same (either both true or both false), and false otherwise.


haithemhamid (changes requested):
you must follow this logic:
If aSmile == true and bSmile == true → trouble → return true
If aSmile == false and bSmile == false → trouble → return true

you need to change the | | becuase it must be && (and)

EchoCoder20 (dismissed):

Ghaida02a (dismissed):
Well done

PR #29, opened 2025-07-08
TRA 566: Sum Double Task Code Added

Ghaida02a:
Implemented a method sumDouble (int a, int b) that checks if the two integers are same it will double their sum. Otherwise, just add the numbers.


EchoCoder20 (changes requested):
Your code logic is correct. But you must return Integer value

haithemhamid (approved):

PR #30, opened 2025-07-08
TRA 568:  Diff 21 task code added

Ghaida02a:
Implement a method diff21(int n) that returns the absolute difference between the given number n and 21.
If n is greater than 21, return double the absolute difference.


EchoCoder20 (dismissed):
Great job
Note: I recommend declaring the global variable inside the function and updating its value based on the required formula for each condition.

haithemhamid (changes requested):
method type must be int because its requirement to return number.

haithemhamid (approved):

EchoCoder20 (approved):

PR #31, opened 2025-07-08
TRA 584: Parrot Trouble code added

Ghaida02a:
Implement a method parrotTrouble(boolean talking, int hour) that returns true if we are in trouble (talking == true)
&& (talking ==true && (hour < 7 || hour > 20))


EchoCoder20 (approved):
Good Job

PR #35, opened 2025-07-08
TRA-609: Monkey Trouble – Determine if We Are in Trouble

Manal-Alhatmi:
This PR adds the method monkeyTrouble, which determines whether we are in trouble based on the smiling status of two monkeys.


haithemhamid (dismissed):

EchoCoder20 (dismissed):
Good!
Note: Next time instead of write `aSmile==false` write `!aSmile`

PR #41, opened 2025-07-08
#tra 639:Not String – Add "not " Prefix Unless Already Present:shatha

sh3tha:
Not String – Add "not " Prefix Unless Already Present


EchoCoder20 (approved):
Well done,
However, I recommend considering all possible inputs. For example, what if the user passes in notString("Not candy")? In this case, "Not" is not equal to "not" because string comparison is case-sensitive. So the method would return "Not candy" instead of "not candy".

PR #42, opened 2025-07-08
#tra 637:Near Hundred – Check if Number is Within 10 of 100 or 200:shatha

sh3tha:
Near Hundred – Check if Number is Within 10 of 100 or 200


EchoCoder20 (approved):
Great,
Note: I recommend assigning the values of equations to separate variables. This way, whenever you need to update the formula, you only have to modify it in one place

PR #43, opened 2025-07-08
TRA-645: Parrot Trouble – Check for Talking Parrot at Bad Hours

Manal-Alhatmi:
This method determines if we are in trouble based on the behavior of a talking parrot.

We are in trouble only if:

The parrot is talking (talking == true), and

The current hour is before 7 AM (hour < 7) or after 8 PM (hour > 20)


EchoCoder20 (changes requested):
You should avoid using if (talking == true) because it’s redundant. The condition inside an if statement is automatically evaluated as a boolean expression. If talking is already a boolean, comparing it to true is unnecessary.

haithemhamid (approved):

PR #48, opened 2025-07-08
TRA-635: Solving Front3 task

EchoCoder20:
Implement a method called front3 that returns a new string made of 3 copies of the front of the string.

The front is defined as the first 3 characters of the string.


haithemhamid (dismissed):
Good but just sure to make good practice of coding by stopping wasting the memory space since the String is immutable.

Ghaida02a (changes requested):
Good job

Ghaida02a (approved):
Well done

PR #49, opened 2025-07-08
#TRA-660: Added Monkey Trouble task

Ghaida02a:
implemented a function monkeyTrouble, that return true we are in trouble if both monkeys are smiling or if non of them is smiling. Otherwise, returns false.


Ghaida02a:
I have modified the code (aSmile && bSmile) || (!aSmile && !bSmile) instead of ==true or ==false

haithemhamid (approved):

EchoCoder20 (approved):

PR #51, opened 2025-07-08
#TRA-606: Added PosNeg file

Ghaida02a:
* Returns true if one of the integers a or b is negative and the other is positive.

* If the negative flag is set to true, the method instead returns true only if both a and b are negative.

* In all other cases, the method returns false.


EchoCoder20 (approved):
Good work

PR #52, opened 2025-07-08
TRA-641: Added Near Hundred 

Ghaida02a:
* Returns true if the given number n is within 10 units (inclusive) of either 100 or 200.

* Utilizes Math.abs() to calculate the absolute difference between n and the target values (100 and 200).


EchoCoder20 (approved):
Great,
Note: I recommend assigning the values of equations to separate variables. This way, whenever you need to update the formula, you only have to modify it in one place

Ghaida02a:
Thank you! That makes sense — assigning the values to variables definitely improves readability and maintainability. I’ll keep that in mind moving forward.

PR #53, opened 2025-07-08
#TRA-622: Added Not String method code

Ghaida02a:
Returns a new string with "not " added to the beginning unless the input string already starts with "not". If the input already begins with "not", the original string is returned unchanged.


Ghaida02a:
Thank you for your comment! That's a great point about case sensitivity. I’ll keep that in mind for future implementations, although for this version I’ll stick with the current logic.

haithemhamid (approved):

PR #54, opened 2025-07-08
#TRA-623: Added Missing Char

Ghaida02a:
* Removes the character at index n from the given string str.

* Returns a new string composed of:
            - Characters before index n: str.substring(0, n)
            - Characters after index n: str.substring(n + 1)

If n is out of bounds (less than 0 or greater than/equal to the string's length), the original string is returned unchanged.


EchoCoder20 (approved):
Well done !

PR #55, opened 2025-07-08
#TRA-624: Added Front Back method

Ghaida02a:
* If the string has 1 character or less, return it unchanged.

* Otherwise:

             - Extract the first and last characters using charAt().
             - Extract the middle portion using substring(1, str.length() - 1).
             - Reconstruct the string as: lastChar + middle + firstChar.


EchoCoder20 (approved):
Great!!!

haithemhamid (approved):

PR #57, opened 2025-07-09
TRA589 This is the Task of SumDouble

kaltham11:
I create a class Sum then I create method call sumDouble
which return int and have two parameter that are a and b 
if a not equal to b then addition a +b
otherwise addition a +b then multiplication with 2


haithemhamid (approved):
just make sure the type of file is java source file

EchoCoder20 (approved):
Note: It is better to declare and initialize the variable sum inside the function.

PR #59, opened 2025-07-09
TRA-654: Missing Char – Remove Character at Given Index

Manal-Alhatmi:
To return a new string formed by removing the character at the specified index n from the input string str.

The input string str is guaranteed to be non-empty.

The index n is always valid, meaning it satisfies: 0 ≤ n < str.length().


haithemhamid (approved):

EchoCoder20 (changes requested):
You need to check that the value of "n" must be more that or equal zero and less than string length. Also you need to check that the string is not empty.

EchoCoder20 (approved):

PR #68, opened 2025-07-09
TRA691: Near Hundred – Check if Number is Within 10 of 100 or 200

haithemhamid:
Implement a method nearHundred(int n) that returns true if the given number n is within 10 of either 100 or 200:
Return true if Math.abs(n - 100) <= 10 or Math.abs(n - 200) <= 10


EchoCoder20 (approved):
I recommend that next time you assign the result of the formula to a separate variable .

NASSER1221 (approved):
good

PR #71, opened 2025-07-09
TRA692: Not String – Add "not " Prefix Unless Already Present

haithemhamid:
Implement a method notString(String str) that returns a new string where "not " is added to the front.
If the original string already starts with "not", return it unchanged.


EchoCoder20 (approved):
Great work!!
Notes for next time:

- You must think about what if the string length is less than 3
- What if `not` word written like this `NOT`??

NASSER1221 (approved):
good

PR #72, opened 2025-07-09
TRA693: Missing Char – Remove Character at Given Index

haithemhamid:
Implement a method missingChar(String str, int n) that returns a new string where the character at index n has been removed.
But the code should aware about:
The input string is non-empty.
The index n will always be valid (0 ≤ n < str.length()).


EchoCoder20 (commented):
why  `if( n==str.lenght())` is invalid index ?

NASSER1221 (dismissed):
very good

haithemhamid:
> why `if( n==str.lenght())` is invalid index ?

because how the length of string can be an index?

PR #74, opened 2025-07-09
TRA694: Front Back – Swap First and Last Characters of a String

haithemhamid:
Implement a method frontBack(String str) that returns a new string where the first and last characters have been swapped.
the code should aware about:
If the string length is 1 or less, return the string unchanged.


NASSER1221 (approved):
Perfect

PR #75, opened 2025-07-09
TRA-698: Solving Or35 Task

EchoCoder20:
Implement a method called or35 which has one param and returns true if the given non-negative number n is a multiple of 3 or 5.


haithemhamid (approved):

Ghaida02a (changes requested):
Good job

PR #76, opened 2025-07-09
TRA695: Front3 – Repeat the Front of a String Three Times

haithemhamid:
Implement a method front3(String str) that returns a new string made of 3 copies of the front of the string.
note:
The front is defined as the first 3 characters of the string.


EchoCoder20 (changes requested):
Please check that If the string length is less than 3, use the whole string as the front.
which mean that if the str="a" the output must be "aaa"

NASSER1221 (approved):
good

EchoCoder20 (approved):

PR #83, opened 2025-07-09
TRA731: StartHi – Check if String Starts with "hi"

haithemhamid:
Implement a method startHi(String str) that returns true if the given string starts with "hi", and false otherwise.


NASSER1221 (approved):
Good

PR #85, opened 2025-07-09
TRA733: In1020 – Check if Either Number is in Range 10 to 20

haithemhamid:
Implement a method in1020(int a, int b) that returns true if either a or b is within the range 10 to 20 (inclusive).


EchoCoder20 (approved):

NASSER1221 (approved):
perfect

PR #86, opened 2025-07-09
TRA734: Has Teen – Check if Any Number is a Teen (13 to 19)

haithemhamid:
Implement a method hasTeen(int a, int b, int c) that returns true if one or more of the given numbers are in the teen range (13 to 19 inclusive).


EchoCoder20 (approved):

NASSER1221 (approved):
Good

PR #87, opened 2025-07-09
TRA-711: Solving last digit Task

EchoCoder20:
Implement a method called lastDigit that pass two integer param  and returns true if the two non-negative integers have the same last digit.


haithemhamid (approved):

Ghaida02a (approved):
Well done

PR #88, opened 2025-07-09
TRA735: Lone Teen – Check if Exactly One Number is a Teen (13 to 19)

haithemhamid:
Implement a method loneTeen(int a, int b) that returns true if exactly one of the two given numbers is in the teen range (13 to 19 inclusive).


EchoCoder20 (approved):

NASSER1221 (approved):
Good

PR #91, opened 2025-07-09
#TRA-625: Added Front3 task code

Ghaida02a:
This pull request adds the implementation of the front3(String str) method, which returns a new string made of 3 copies of the first 3 characters of the input string.


EchoCoder20 (changes requested):
Please check the condition if(str.length() < 3). The method should use the entire string as the front. For example, if str = 'a', the method must return 'aaa' rather than just 'a'. Make sure to modify the logic to handle cases like this properly."

haithemhamid (changes requested):

haithemhamid (approved):

PR #92, opened 2025-07-09
#TRA-655: Back Around Added

Ghaida02a:
This pull request introduces the backAround(String str) method, which returns a new string where the last character of the original string is added to both the front and back.

* If the input string has at least one character:
       - Extract the last character using charAt(str.length() - 1).
       - Construct a new string as: lastChar + str + lastChar.


EchoCoder20 (approved):
Well done!!
Note: You can avoid use `else` block and only write `return ""; directly after `if`.

haithemhamid (approved):

PR #95, opened 2025-07-09
TRA-619: Add Near Hundred task

Hawraa-a:
I implemented a method nearHundred(int n) that checks if a number within 10 of 100 or 200 using Math.abs() to compute the absolute difference. It returns true if the condition is met.


EchoCoder20 (approved):
Well done!
Note for next time: next time you need to assign the value of the formula to a variable so next time if the formula change we can directly change it in one line.

PR #99, opened 2025-07-09
TRA749: IntMax – Return the Largest of Three Integers

haithemhamid:
Implement a method intMax(int a, int b, int c) that returns the largest of the three given integer values.


EchoCoder20 (approved):

NASSER1221 (approved):
Good

PR #102, opened 2025-07-09
TRA752: In3050 – Check if Both Numbers Are in 30..40 or 40..50 Range

haithemhamid:
Implement a method in3050(int a, int b) that returns true if both numbers are:
In the range 30 to 40 (inclusive) or in the range 40 to 50 (inclusive).


EchoCoder20 (approved):

NASSER1221 (approved):
Good

PR #103, opened 2025-07-09
TRA753: StringE – Check if String Contains Between 1 and 3 'e' Characters

haithemhamid:
Implement a method stringE(String str) that returns true if the string contains between 1 and 3 'e' characters (inclusive).

### DIFF
--- a/plag_test_tool/from_Haitham/TRA753.java
+++ b/plag_test_tool/from_Haitham/TRA753.java
@@ -1,0 +1,13 @@
+public class TRA753{
+public static boolean stringE(String str) {
+        char[] subStrings = new char[3];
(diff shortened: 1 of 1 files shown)


EchoCoder20 (changes requested):
Good attempt, but I think there's a misunderstanding of the question. You need to count how many times the character 'e' appears in the string. If it appears more than 3 times, return false; otherwise, return true.

Please test your code with the following cases:

stringE("Hello") → true
stringE("Heelle") → true
stringE("Heelele") → false"

NASSER1221 (approved):
Good

EchoCoder20 (approved):

PR #104, opened 2025-07-09
TRA754: Max1020 – Return Larger Value in Range 10..20 or 0 if None

haithemhamid:


EchoCoder20 (approved):
Good job!!
Note for next time : Avoid use extra parentheses around each condition

PR #105, opened 2025-07-09
TRA755: Last Digit – Check if Two Numbers Have the Same Last Digit

haithemhamid:
Implement a method lastDigit(int a, int b) that returns true if the two non-negative integers have the same last digit.
By using the modulo operator % to get the last digit (number % 10).


EchoCoder20 (approved):

NASSER1221 (approved):
good

PR #106, opened 2025-07-09
TRA756: EndUp – Uppercase Last 3 Characters of a String

haithemhamid:
Implement a method endUp(String str) that returns a new string where the last 3 characters are converted to uppercase.
With make sure of that:
If the string length is less than 3, uppercase the whole string.


EchoCoder20 (changes requested):
It is better to check if the string is empty before checking its length, because if the string is empty the first condition actually will executed not the second condition.

EchoCoder20 (approved):

PR #111, opened 2025-07-09
TRA-706:  Front Back – Swap First and Last Characters of a String

Manal-Alhatmi:
I Implemented a method frontBack(String str) that returns a new string where the first and last characters have been swapped.

If the string length is 1 or less, return the string unchanged.


haithemhamid (changes requested):
you need to swap First and Last Characters of a String.

EchoCoder20 (dismissed):
Good !
Next time consider the condition if the string is empty

PR #115, opened 2025-07-10
TRA-651: Added Front Back task

Hawraa-a:
I implemented a method called frontBack(String str) that takes a string as input and returns a new string where the first and last characters are swapped. If the string has a length of 1 or less, the method returns it unchanged.


EchoCoder20 (approved):
Great work!

haithemhamid (approved):

PR #116, opened 2025-07-10
#TRA-657: Front22 task code added

Ghaida02a:
This pull request adds the implementation of the front22(String str) method, which returns a new string with the first 2 characters added to both the front and back of the original string.

* If the input string has fewer than 2 characters:
      - Return the string repeated 3 times.
* Otherwise:
      - Extract the first 2 characters using substring(0, 2).
      - Return the result of: first2 + str + first2.


EchoCoder20 (approved):
Good thinking!

haithemhamid (approved):

PR #119, opened 2025-07-10
#TRA-700: Icy Hot task code added

Ghaida02a:
This pull request adds the implementation of the icyHot(int temp1, int temp2) method, which checks if one temperature is below 0 and the other is above 100.
* Returns true if:
     - One of the temperatures is less than 0 (icy) && the other is greater than 100 (hot)
* Returns false otherwise.


EchoCoder20 (approved):
Good,
Note: You can avoid used multiple of  parentheses `()` inside if statement.

haithemhamid (approved):

PR #121, opened 2025-07-10
TRA-661: Added Back Around task

Hawraa-a:
I implemented a method backAround(String str). It returns a new string where the last character of the original string is added to both the beginning and end of the string.


EchoCoder20 (approved):
Good!!

haithemhamid (approved):

PR #122, opened 2025-07-10
#TRA-701: In1020 task code added

Ghaida02a:
This pull request adds the implementation of the in1020(int a, int b) method, which checks whether either of the two integers is within the range 10 to 20 (inclusive).

* Returns true if either a or b is between 10 and 20, inclusive.
* Returns false otherwise.


EchoCoder20 (approved):
Well done!!

haithemhamid (approved):

PR #127, opened 2025-07-10
TRA-710:  Front22 – Add First Two Chars to Front and Back of String

Manal-Alhatmi:
I Implemented a method front22(String str) that returns a new string with the first two characters added to both the front and back of the original string.

If the string length is less than 2, use the whole string as the prefix and suffix.


haithemhamid (approved):
```
 if (n > str.length()) {
            n =str.length();
        }
```
no need to handle the case of n> length of string as above.

PR #128, opened 2025-07-10
TRA-779: Solving IcyHot Task

EchoCoder20:
Implement a method called icyHot that pass two integer param and returns true if one temperature is less than 0 and the other is greater than 100.


haithemhamid (changes requested):
you need to cover all cases for this task e.g. temp1=-10, temp2=35.

PR #129, opened 2025-07-10
TRA-662: Added Or35 task

Hawraa-a:
In this task, I created a method or35(), that takes an integer "n" as input and returns a Boolean value. Inside the method, I checked two conditions: 
- Whether the number is greater than 0. 
- Whether it is divisible by 3 or by 5 using the modulo operator (%).
If both conditions are true, the method returns true. Otherwise, it returns false.


haithemhamid (approved):
just you need to be sure if consider n to greater than 0 then the code cannot handle the negative input but if we try with n=-10 it will return true.

PR #132, opened 2025-07-10
TRA-614

Mulk000:
I wrote a Java program that defines a method to add two integers, but with a special rule:
*If the two numbers are equal, return double their sum.
Otherwise, return their regular sum.


haithemhamid (approved):
same note about .txt extension

PR #148, opened 2025-07-10
TRA-672

Mulk000:
I made a program that removes one letter from a word at a given position.
I give it a word (string) and a number (index) and it removes the letter at that index.


EchoCoder20 (changes requested):
You should check the value of n to ensure it satisfies the condition 0 ≤ n < str.length(). Additionally, make sure that the input string is not empty before performing any operations.

EchoCoder20 (approved):

PR #158, opened 2025-07-10
TRA-771

Mulk000:
I made a program that checks if a number is a multiple of 3 or 5.
*If the number can be divided evenly by 3 or 5, return true.
*Otherwise, return false.


haithemhamid (approved):

EchoCoder20 (changes requested):
Good work!
But you need to check the value of 'n' must not negative.

PR #161, opened 2025-07-10
Monkey Trouble – Determine if We Are in Trouble

hanin190340:
Implement a method `monkeyTrouble(boolean aSmile, boolean bSmile)` that determines if we are in trouble based on whether two monkeys are smiling. We are considered to be in trouble if **both monkeys are smiling** or if **neither monkey is smiling**—that is, when `aSmile` and `bSmile` are either both `true` or both `false`. In all other cases, we are **not** in trouble and the method should return `false`. This task is designed to help practice boolean logic and the use of logical operators. For example, `monkeyTrouble(true, true)` returns `true`, `monkeyTrouble(false, false)` returns `true`, and `monkeyTrouble(true, false)` returns `false`.


EchoCoder20 (approved):
Great

haithemhamid (approved):

PR #163, opened 2025-07-10
#TRA-703: Lone Teen task added

Ghaida02a:
This pull request implements the loneTeen(int a, int b) method, which returns true if exactly one of the two input values is in the "teen" range (13 through 19 inclusive).

*Checks whether only one of the two integers (a or b) is a "teen", between 13 and 19

        Returns:
            - true if exactly one of the values is a teen.
            - false if both are teens or neither is a teen.


EchoCoder20 (approved):
Good job.!!

haithemhamid (approved):

PR #171, opened 2025-07-10
TRA-665: Added Icy Hot task

Hawraa-a:
I implemented the icyHot() method, which returns true if one of the input temperatures is less than "0" and the other is greater than "100".


EchoCoder20 (approved):
Good

haithemhamid (approved):

PR #172, opened 2025-07-10
#TRA-762: IntMax task added

Ghaida02a:
This pull request introduces the intMax(int a, int b, int c) method, which returns the largest of three integer values.

*Compares three integers: a, b, and c.
*Returns the one with the highest value:
         - If a is greater than or equal to both b and c, return a.
         - Else if b is greater than or equal to both a and c, return b.
         - Otherwise, return c


EchoCoder20 (approved):
Well done !

haithemhamid (approved):

PR #176, opened 2025-07-10
#TRA-763: Close10 tak added

Ghaida02a:
This pull request adds the close10(int a, int b) method, which determines which of the two input integers is closer to the value 10. If both are equally close, it returns 0.

*Computes the absolute difference between each number and 10:
        - A = Math.abs(a - 10)
        - B = Math.abs(b - 10)
* Returns:
- a if it is closer to 10.
- b if it is closer to 10.
- 0 if both are equally close to 10.


EchoCoder20 (approved):
Great

PR #178, opened 2025-07-10
#TRA-764: In3050 task added

Ghaida02a:
This pull request introduces the in3050(int a, int b) method, which checks whether both integers fall within the same valid range: either 30–40 or 40–50, inclusive.


EchoCoder20 (approved):
Good work!!

haithemhamid (approved):

PR #180, opened 2025-07-10
TRA811: Makes 10 – Check if Either Value or Their Sum is 10

hanin190340:
Implement a method makes10(int a, int b) that returns true if either of the integers a or b is equal to 10, or if their sum is 10. If neither condition is met, the method should return false. This task is intended to practice using logical OR conditions (||) and basic arithmetic comparisons. For example, makes10(9, 10) returns true, makes10(9, 9) returns false, and makes10(1, 9) returns true.


haithemhamid (approved):

Ghaida02a (approved):
Good job

from_hanin/makes10.txt
```diff
@@ -0,0 +1,6 @@
+public class Makes10 {
```
EchoCoder20, line 1: Class name must match the filename

Ghaida02a: Also, Class violates naming conventions.

EchoCoder20 (approved):

PR #182, opened 2025-07-10
#TRA-766: StringE task added

Ghaida02a:
This pull request adds the stringE(String str) method, which checks whether the given string contains between 1 and 3 occurrences of the character 'e' (inclusive).
*Iterates over each character in the string.
*Counts how many times the character 'e' appears.
*Returns true if the count is between 1 and 3, inclusive.
*Returns false if the count is 0 or more than 3.


EchoCoder20 (approved):
Great

haithemhamid (approved):

PR #186, opened 2025-07-10
TRA: PosNeg – Return True Based on Value Signs and a Flag

hanin190340:
Implement a method posNeg(int a, int b, boolean negative) that returns true based on the sign of the integers a and b, with behavior influenced by the negative flag. If negative is false, return true only if one value is negative and the other is positive. If negative is true, return true only if both values are negative. This task focuses on practicing conditional evaluation using boolean flags and sign checks. For example, posNeg(1, -1, false) returns true, posNeg(-1, 1, false) returns true, and posNeg(-4, -5, true) returns true.

### DIFF
--- a/PosNeg.txt
+++ b/PosNeg.txt
@@ -1,14 +1,14 @@
 public class PosNeg {
     public static boolean posNeg(int a, int b, boolean negative)
     {
-        if( a<0 && b<0 && negative==true){
+        if( a<0 && b<0 && negative){
             return true;
 
-        } else if (a>0 && b<0 && negative==false) {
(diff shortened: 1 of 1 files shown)


EchoCoder20 (changes requested):
Your logic is correct, but instead of write the expression inside if statment like this : `if(negative==True)` you need to write if like this `if(negative)` beacuse `if(negative==True)`  is equle to` if(negative)` and` if(negative==false)` is same as `if(!negative).`

EchoCoder20 (approved):

haithemhamid (approved):

PR #187, opened 2025-07-10
#TRA-769: EndUp task added

Ghaida02a:
This pull request adds the endUp(String str) method, which returns a version of the input string where the last 3 characters are converted to uppercase. If the string has fewer than 3 characters, the entire string is converted to uppercase.

*If the string has less than 3 characters, return the entire string in uppercase.
*Otherwise:
     - Split the string into two parts:
          - lowerCase: all characters except the last 3.
          - upperCase: the last 3 characters.
     - Return the concatenation of lowerCase + upperCase.toUpperCase().


EchoCoder20 (approved):
Good job

PR #188, opened 2025-07-10
#TRA-770: EveryNth task added

Ghaida02a:
This pull request implements the everyNth(String str, int n) method, which returns a new string made by selecting every nth character from the input string, starting with the first character (index 0).
* Iterates over the input string in steps of n.
* Appends the character at each step to a StringBuilder.
* Returns the constructed string containing every nth character.


EchoCoder20 (changes requested):
Good logic but ,you need to check  the value of `n` if it is more than or equal to (1) otherwise we can return "Invalid (n) Value"

Ghaida02a:
I have modified the code. Added if condition (if n is less than 1, return "Invalid (n) Value")

haithemhamid (changes requested):
please make sure of these two conditions if satisfied:
n is guaranteed to be 1 or more.
The input string is non-empty.

EchoCoder20 (dismissed):

EchoCoder20 (approved):

haithemhamid (approved):

PR #190, opened 2025-07-10
TRA813:Near Hundred – Check if Number is Within 10 of 100 or 200

hanin190340:
Implement a method `nearHundred(int n)` that returns `true` if the number `n` is within 10 units of either 100 or 200. To determine this, use `Math.abs()` to calculate the absolute difference between `n` and 100, and between `n` and 200. Return `true` if either difference is less than or equal to 10; otherwise, return `false`. This exercise helps practice using `Math.abs()` and writing conditions based on numerical proximity. For example, `nearHundred(93)` and `nearHundred(90)` return `true`, while `nearHundred(89)` returns `false`.

### DIFF
--- a/NearHundred.txt
+++ b/NearHundred.txt
@@ -1,0 +1,12 @@
+public class NearHundred {
+    public static boolean  nearHundred(int n)
+    {
+        if (Math.abs(n - 100) <= 10 || Math.abs(n - 200) <= 10)
+        {
+            return true;
+        }
+        else {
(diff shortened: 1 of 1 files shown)


EchoCoder20 (approved):
Great work!!
Note for next time: I recommended to assign the value of the formula in separate variable so, when you want to change the formula you change in one place.

haithemhamid (approved):

PR #193, opened 2025-07-10
TRA814:Not String – Add "not " Prefix Unless Already Present

hanin190340:
Implement a method `notString(String str)` that returns a new string with `"not "` added to the front unless the original string already starts with `"not"`. If the string begins with `"not"`, return it unchanged. Use `.equals()` or `.startsWith()` for accurate string comparison. This task helps practice string manipulation and conditional logic. For example, `notString("candy")` returns `"not candy"`, `notString("x")` returns `"not x"`, and `notString("not bad")` returns `"not bad"`.

### DIFF
--- a/Not String.txt
+++ b/Not String.txt
@@ -1,0 +1,10 @@
+public class NotString {
+    public static String notString(String str) {
+        if (str.equals("not") || str.startsWith("not")) {
+            return str ;
+        }
+         else  {
+          return "not "+ str;
+        }
+    }
+}


EchoCoder20 (approved):
Well done!
However, I recommend considering all possible inputs. For example, what if the user passes in notString("Not candy")? In this case, "Not" is not equal to "not" because string comparison is case-sensitive. So the method would return "Not candy" instead of "not candy".
Think about it

PR #197, opened 2025-07-10
TRA-740: Added In3050 task

Hawraa-a:
I implemented the in3050(int a, int b) method, which returns true if both input numbers are either in the range 30 to 40 (inclusive) or both are in the range 40 to 50 (inclusive).


EchoCoder20 (approved):
Good works

haithemhamid (approved):

PR #200, opened 2025-07-10
TRA-743: Added Last Digit task

Hawraa-a:
I implemented the lastDigit(int a, int b) method, which returns true if the two non-negative integers have the same last digit. The method uses the modulo operator % to compare the last digits.


EchoCoder20 (approved):
Well done!

haithemhamid (approved):

PR #202, opened 2025-07-10
TRA-745: Added EveryNth task

Hawraa-a:
I implemented the everyNth(String str, int n) method, which returns a new string made of characters starting at index 0 and then every nth character thereafter. The input string is guaranteed to be non-empty, and n is always 1 or greater.


EchoCoder20 (changes requested):
You need to check the value of "n" ,it must be more than or equal 1  ("n is guaranteed to be 1 or more.")  then you can use the` for loop ` ,otherwise you can return meaningful message such as "Invalid value of N"

EchoCoder20 (changes requested):
You need to check the value of the `n` parameter not the length of string

EchoCoder20 (approved):

haithemhamid (approved):

PR #203, opened 2025-07-10
TRA513: SleepIn

Alfori90:
The goal of this task is to implement the `sleepIn(boolean weekday, boolean vacation)` method, which determines whether we can sleep in based on the day and vacation status. According to the logic, we are allowed to sleep in if it is **not a weekday** (`weekday == false`) or if we are **on vacation** (`vacation == true`). The method should return `true` in either of these situations, and `false` otherwise. This exercise helps practice basic conditional statements and the use of boolean operators in decision-making.

### DIFF
--- a/Huria/Sleep-In Logic – Implement sleepIn(weekday, vacation).txt
+++ b/Huria/Sleep-In Logic – Implement sleepIn(weekday, vacation).txt
@@ -1,0 +1,30 @@
+public class SleepIn {
+
+    boolean weekday;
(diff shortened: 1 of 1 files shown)


EchoCoder20 (changes requested):
Nice try, but the logic in the `if` condition needs adjustment. It should return `true` if it's not a weekday or if there is a vacation. Otherwise, it should return `false`."

haithemhamid (approved):

PR #206, opened 2025-07-10
TRA514:Sum Double

Alfori90:
The objective of this task is to implement the sumDouble(int a, int b) method, which returns the sum of two integers. If the two integers are different, the method should simply return their sum (a + b). However, if the two integers are equal, it should return double their sum (2 * (a + b)). This exercise helps reinforce the use of conditional statements and basic arithmetic operations in programming.


EchoCoder20 (approved):
Good job,
Note for next time: I recommend you to move the calculations inside the if and else blocks to avoid unnecessary computation. Also, you can only declare or define one variable that hold the value of the calculation .

PR #208, opened 2025-07-10
TRA517: Parrot Trouble

Alfori90:
The goal is to implement the method parrotTrouble(boolean talking, int hour), which returns true if we are in trouble based on the parrot’s behavior and the time of day. We are considered in trouble if the parrot is talking (talking == true) and the current hour is either before 7 (hour < 7) or after 20 (hour > 20). In all other cases, the method should return false. This exercise focuses on practicing how to combine boolean and comparison operators in conditional statements.


EchoCoder20 (dismissed):
Great work!

PR #211, opened 2025-07-12
TRA600: Back Around – Add Last Char to Front and Back of String

kaltham11:
I create a class called BackAroundString then I create method called backAround
which return String and have one parameter that is String (str)  
the method check if the lenght of str is not equal to 0 if contion true
then converte str to stringBilder(newStr) 
take the last index and save it in char (lastIndex)
then I used the method insert to add the char (lastIndex) in 0 index
and again insert char (lastIndex) in last index 
then convers (newStr)to (finalString)
then return finalString
otherwise return "The string is null"

I created a class called BackAroundString
In the method backAround, I first check if the string is not empty.
Then I convert the string to a char array named strToArray.
I create a new char array that is 2 characters longer than the original.
I store the last character of the original string in a variable lastIndex.
I place lastIndex at the beginning and end of the new array.
I fill the rest of the new array with the characters from the original string.
Finally, I convert the new array to a String and return it.
If the string is empty, I return "The string is null".

### DIFF
--- a/from_Kaltham/BackAround.txt
+++ b/from_Kaltham/BackAround.txt
@@ -1,0 +1,37 @@
+//This is first solution:
+public class BackAroundString {
+    public String backAround(String str) {
+        if (str.length() != 0) {
+            StringBuilder newStr = new StringBuilder(str);
+            char lastIndex = newStr.charAt(newStr.length() - 1);
+            newStr.insert(0,lastIndex);
+            newStr.insert(newStr.length(),lastIndex);
+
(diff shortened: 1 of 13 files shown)


haithemhamid (approved):

kaltham11:
I used CharAt () in first solution, I didn't converted string to array of char. I converted to StringBuilder because I want to Change the string and the string is immutable.

The second solution I converted string to array of char. and that is another solution. I didn't use CharAt() in that solution.

PR #213, opened 2025-07-12
TRA796: Front22 – Add First Two Chars to Front and Back of String

kaltham11:
//The First description For First code:
I created a class called FrontEndString and the method front22, I extract the first two characters of the input string using substring(0, 2) and store it in str1. Then I return a new string by adding str1 before and after the original string.
So the result is the first two characters + the original string + the first two characters again.

//The Second description For second code:
I created a class called FrontEndString and the method front22, I first check if the string is not empty.
Then I convert the string into a char array called strToArray.
I create a new char array called newchars that is 4 characters longer than the original string.
I create another array firstTwoIndex that stores the first two characters of the string.
I copy firstTwoIndex into the beginning of newchars using a for-each loop.
Then I copy the original string characters into newchars starting from index 2.
After that, I copy firstTwoIndex again to the end of newchars.
Finally, I convert the new array to a String and return it.
If the string is empty, I return "The string is null".

//The Third description For Third code:
I created a class called FrontEndString.
In the method front22, I first check if the input string is not empty.
I convert the string into a char array named strToArray.
I prepare a new char array called newchars, which is 4 characters longer than the original.
I use Arrays.copyOfRange to extract the first two characters and store them in firstTwoIndex.
Using System.arraycopy, I copy:
firstTwoIndex to the beginning of newchars,
the full strToArray right after that,
and again firstTwoIndex at the end.
Finally, I convert the new array back to a String and return it.
If the string is empty, I return "The string is null".

### DIFF
--- a/from_Kaltham/BackAround.txt
+++ b/from_Kaltham/BackAround.txt
@@ -1,0 +1,37 @@
+//This is first solution:
+public class BackAroundString {
+    public String backAround(String str) {
+        if (str.length() != 0) {
+            StringBuilder newStr = new StringBuilder(str);
+            char lastIndex = newStr.charAt(newStr.length() - 1);
+            newStr.insert(0,lastIndex);
+            newStr.insert(newStr.length(),lastIndex);
+
+            String finalString= newStr.toString();
+            return finalString;
+        }
+        return "The string is null";
+    }
+}
+
+//This is second solution:
+
+public class BackAroundString {
+    public String backAround(String str) {
+        if (str.length() != 0) {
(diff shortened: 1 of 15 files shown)


haithemhamid (approved):

EchoCoder20 (changes requested):
You need to consider this condition " If the string length is less than 2, use the whole string as the prefix and suffix." in your code logic

EchoCoder20 (approved):
Note: Next time you need to consider "if the string is empty or not"

PR #219, opened 2025-07-12
TRA802: DelDel – Remove "del" Starting at Index 1 if Present

kaltham11:
//The first description for first code:
I created a class called DelDel.
In the method delDel, I check if the substring from index 1 to 4 is equal to "del".
If it is, I create a StringBuilder from the original string.
I use the delete(1, 4) method to remove the characters from index 1 to 3 ("del").
Then I convert the StringBuilder back to a string and return the result.
If "del" is not found at index 1, I simply return the original string.

//The second description for second code:
I created a class called DelDel. In the method delDel, I first convert the input string into a char array called chars.
I prepare a new array newChars that is 3 characters shorter than the original.
Then I check if the characters at index 1, 2, and 3 are 'd', 'e', and 'l' respectively.
If so, I use a for loop to copy all characters from the original array except indices 1, 2, and 3.
I fill newChars with the remaining characters and return it as a new string.
If "del" is not found starting at index 1, I return the original string.

### DIFF
--- a/from_Kaltham/BackAround.txt
+++ b/from_Kaltham/BackAround.txt
@@ -1,0 +1,37 @@
+//This is first solution:
+public class BackAroundString {
+    public String backAround(String str) {
+        if (str.length() != 0) {
+            StringBuilder newStr = new StringBuilder(str);
+            char lastIndex = newStr.charAt(newStr.length() - 1);
+            newStr.insert(0,lastIndex);
+            newStr.insert(newStr.length(),lastIndex);
+
+            String finalString= newStr.toString();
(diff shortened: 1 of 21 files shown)


haithemhamid (approved):

EchoCoder20 (changes requested):
Note that If the input string is less than 4 characters, str.substring(1, 4) will cause a StringIndexOutOfBoundsException, so you need to check the length of the string (is must be more than or equal 4) and check if the string is empty or not.

EchoCoder20 (approved):

PR #226, opened 2025-07-12
TRA-823: Solving Close10 Task

EchoCoder20:
Implement a method close10(int a, int b) that returns the value which is nearest to 10.

If both numbers are equally close to 10, return 0.

Use Math.abs() to calculate the absolute difference.


haithemhamid (changes requested):
The file is empty

Ghaida02a (changes requested):

PR #230, opened 2025-07-12
TRA822: EveryNth – Build String Using Every Nth Character

kaltham11:
I created a class called EveryNth. In the method everyNth:
I convert the input string to a StringBuilder object named newStr.
I create another StringBuilder named finalStr to store the result.
I loop through the string from index 0, increasing by n each time:
On each step, I take the character at index i using charAt(i).
I append it to finalStr.
After the loop ends, I return finalStr as a string using .toString().

### DIFF
--- a/from_Kaltham/BackAround.txt
+++ b/from_Kaltham/BackAround.txt
@@ -1,0 +1,37 @@
+//This is first solution:
+public class BackAroundString {
+    public String backAround(String str) {
+        if (str.length() != 0) {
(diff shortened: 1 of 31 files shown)


EchoCoder20 (changes requested):
n is guaranteed to be 1 or more, but in your code you checked only `if n>1`
The input string is non-empty-> to check this you can use `str.isEmpty()` which is more better that` str.length()>1`

EchoCoder20 (approved):

PR #234, opened 2025-07-13
#tra 479: StartHi – Check if String Starts with "hi"

sh3tha:
StartHi – Check if String Starts with "hi"


EchoCoder20 (approved):
Well done!!
Next time ,I recommend considering all possible inputs. For example, what if the user passes in startHi("Hi ")? In this case, "Hi" is not equal to "hi" because string comparison is case-sensitive. So the method would return "false instead of "true".

PR #238, opened 2025-07-13
TRA-828: DelDel – Remove "del" Starting at Index 1 if Present

Manal-Alhatmi:
I Implemented a method delDel(String str) that checks if the substring "del" appears starting at index 1 of the string.

If yes, return a new string with that "del" removed.

Otherwise, return the string unchanged.


EchoCoder20 (approved):
Great!

PR #252, opened 2025-07-13
TRA522: NotString

Alfori90:
Implement a method notString(String str) that returns the original string with "not " added to the front, unless the string already starts with "not", in which case it is returned unchanged. Use .equals() for comparisons to practice string manipulation and conditionals.


EchoCoder20 (approved):
Great work!

PR #253, opened 2025-07-13
TRA508: Diff 21 - Return Absolute Difference or Double If Over 21

Shaimaalnabhani:
The diff21 method returns how far a number is from 21. If the number is over 21, it returns double that difference.


haithemhamid (approved):

EchoCoder20 (approved):
Good!!
Note: It is enough to write `return diff` only one time after line 13 instead of repeat it twice

PR #261, opened 2025-07-13
TRA524:Front Back

Alfori90:
Implement a method front back(String str) that returns a new string with the first and last characters swapped. If the string has a length of 1 or less, return it unchanged.


haithemhamid (approved):

EchoCoder20 (approved):
Well done!

PR #265, opened 2025-07-13
TRA526: Back Around

Alfori90:
I Implement a method back around(String str) that returns a new string with the last character of the original string added to both the beginning and the end. The input string will have a length of at least 1.


EchoCoder20 (approved):
Great!

PR #274, opened 2025-07-13
TRA-847:  Max1020 – Return Larger Value in Range 10..20 or 0 if None

Manal-Alhatmi:
I Implemented a method max1020(int a, int b) that returns the larger value between a and b that is in the range 10 to 20 (inclusive) ,If neither value is in the range, return 0


haithemhamid (changes requested):
```
if (b > a) {
            int t = a;
            a = b;
            b = t;
        }
```
incorrect approach for doing this

PR #275, opened 2025-07-13
#tra 497: EveryNth – Build String Using Every Nth CharSystem.out.println(EndUp.endUp("hi there"));acter:shatha

sh3tha:
EveryNth – Build String Using Every Nth CharSystem.out.println(EndUp.endUp("hi there"));acter


EchoCoder20 (changes requested):
You need to check the value of "n" ,it must be more than or equal 1 ("n is guaranteed to be 1 or more.") then you can use the for loop ,otherwise you can return meaningful message such as "Invalid value of N"

EchoCoder20 (approved):

haithemhamid (approved):

PR #286, opened 2025-07-13
TRA-849:  Last Digit – Check if Two Numbers Have the Same Last Digit

Manal-Alhatmi:
I Implemented a method lastDigit(int a, int b) that returns true if the two non-negative integers have the same last digit.

Use the modulo operator % to get the last digit (number % 10).


haithemhamid (approved):

EchoCoder20 (changes requested):
You need to make sure that the value of "a" and "b" are not negative

PR #289, opened 2025-07-13
TRA535:Lone Teen

Alfori90:
I implement a method loneTeen(int a, int b) that returns true if exactly one of the two numbers is in the teen range (13 to 19 inclusive); otherwise, it returns false.


EchoCoder20 (approved):
Well done!

PR #290, opened 2025-07-13
Parrot Trouble – Check for Talking Parrot at Bad Hours

NASSER1221:
* Practiced data types and conditions.


EchoCoder20 (changes requested):
Good work!
But change `if (talking)` instead of `if(talking==true )`

PR #291, opened 2025-07-13
TRA-850: EndUp – Uppercase Last 3 Characters of a String

Manal-Alhatmi:
I Implemented a method endUp(String str) that returns a new string where the last 3 characters are converted to uppercase.
If the string length is less than 3, uppercase the whole string.


EchoCoder20 (approved):
Good!
Just avoid use unnecessary Curly brackets

PR #292, opened 2025-07-13
TRA861: Missing Char – Remove Character at Given Index

hanin190340:
The `missingChar` method solves the task by using string slicing to remove the character at the specified index `n` from the input string. It does this by combining two parts of the string: the substring from the beginning up to (but not including) index `n`, and the substring from index `n + 1` to the end. This effectively skips over the character at position `n`, resulting in a new string with that character removed. The approach is efficient and leverages built-in string operations, ensuring the solution is concise, readable, and meets the problem constraints.

### DIFF
--- a/Missing Char.txt
+++ b/Missing Char.txt
@@ -1,0 +1,12 @@
+public class MissingChar {
+    public static String missChar (String str, int n)
+    {
+        if (str != null && 0 <= n && n < str.length()) {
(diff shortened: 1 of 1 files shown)


EchoCoder20 (approved):
Good work!
Note for next time: I recommended if condition inside `if` statement return false, to return meaningful message instead of return `str`

haithemhamid (approved):

PR #298, opened 2025-07-13
TRA867:Front22 – Add First Two Chars to Front and Back of String

hanin190340:
The front22 method creates a new string by taking the first two characters of the given string and appending them to both the beginning and end of the original string. If the input string has fewer than two characters, the entire string is used instead of just the first two. This is achieved by slicing the string using str.substring(0, Math.min(2, str.length())) to safely extract the correct prefix regardless of the string's length. The method then concatenates this prefix + original string + prefix to form the final result, reinforcing the practice of string slicing and concatenation.

### DIFF
--- a/Front22.txt
+++ b/Front22.txt
@@ -1,0 +1,13 @@
+public class Front22 {
+    public static String front22(String str)
+    {
+        if(str.length()<2 ){
+            return str;
+        }
+        else {
+            StringBuilder newStr =new StringBuilder(str);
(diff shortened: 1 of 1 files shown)


EchoCoder20 (changes requested):
Please follow the logic if the string is less than 2 use the whole string as the prefix and suffix.
which mean front22("a") it return → "aaa"

EchoCoder20 (approved):

PR #300, opened 2025-07-13
TRA869: 🌡️Icy Hot – Check Temperatures for Opposite Extremes

hanin190340:
The icyHot method checks two integer temperature values and returns true only if one of them is below 0 (icy) and the other is above 100 (hot). This is achieved using a compound boolean condition that evaluates whether temp1 < 0 && temp2 > 100 or temp2 < 0 && temp1 > 100. The method ensures that the extremes are opposite, practicing logical operators and conditional evaluation — a key concept in control flow and decision-making.

### DIFF
--- a/Icy Hot.txt
+++ b/Icy Hot.txt
@@ -1,0 +1,12 @@
+public class IcyHot {
+    public static boolean icyHot(int temp1, int temp2)
+    {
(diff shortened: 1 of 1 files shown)


EchoCoder20 (approved):
Good

haithemhamid (approved):

PR #323, opened 2025-07-13
TRA879:Max1020 – Return Larger Value in Range 10..20 or 0 if None

hanin190340:
The max1020 method checks two integers and returns the larger one that falls within the inclusive range 10 to 20. If both numbers are within the range, it returns the maximum of the two. If only one number is within the range, it returns that number. If neither number is in the range, it returns 0. This method helps practice range validation and conditional logic to select values based on constraints.

### DIFF
--- a/Max1020.txt
+++ b/Max1020.txt
@@ -1,0 +1,12 @@
+public class max1020 {
+    public static int max1020(int a, int b) {
+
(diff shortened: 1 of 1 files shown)


EchoCoder20 (approved):
Good,
Note: In line 6, you can directly use `else` block instead of using `else if`

PR #331, opened 2025-07-13
Not String – Add "not " Prefix Unless Already Present

NASSER1221:
*Practiced string manipulation and conditions.


EchoCoder20 (approved):
Good!
Next time try to validate that the string is not empty . Also, "not" is not same like "Not" and "NOT" and so on in java ,so you need next time to consider all possible input. To deal with that you can converting the input string to lowercase using str.toLowerCase().startsWith("not") or even to UpperCase.

PR #335, opened 2025-07-13
TRA883: EveryNth – Build String Using Every Nth 

hanin190340:
The everyNth method constructs a new string by starting from the first character (index 0) of the input string and then selecting every nth character thereafter. By looping through the string with increments of n, the method efficiently extracts characters at the desired positions, practicing controlled string iteration and indexing.


EchoCoder20 (approved):
Good,
Note: You do not need to use two `stringBuilders` its enough to use only one because string builder is not immutable.

haithemhamid (approved):

PR #337, opened 2025-07-13
TRA545: Last Digit add

Alfori90:
I implement a method lastDigit(int a, int b) that returns true if the two non-negative integers have the same last digit. Use the modulo operator % (i.e., a % 10 and b % 10) to extract the last digits.


haithemhamid (approved):
you need to check if the given input is:
The two non-negative integers.

EchoCoder20 (dismissed):

PR #339, opened 2025-07-13
Missing Char – Remove Character at Given Index

NASSER1221:
*Practiced indexing and conditioning.


EchoCoder20 (changes requested):
Good!
But you need validate that the string is not empty and the value of "n" must not more than the string length and not less than 0

NASSER1221:
Added the condition.

PR #341, opened 2025-07-13
TRA546: EndUp Add

Alfori90:
I implement a method endUp(String str) that returns a new string with the last 3 characters converted to uppercase. If the string has fewer than 3 characters, convert the entire string to uppercase.

### DIFF
--- a/Huria/EndUp – Uppercase Last 3 Characters of a String.txt
+++ b/Huria/EndUp – Uppercase Last 3 Characters of a String.txt
@@ -8,12 +8,12 @@ public class EndUp {
                 return str.toUpperCase();
             }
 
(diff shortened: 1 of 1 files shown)


Huria/EndUp – Uppercase Last 3 Characters of a String.txt
```diff
@@ -0,0 +1,29 @@
...
+        String str;
+
+        public static String endUp(String str) {
+
+            if (str.length()<=3){
+                return str.toUpperCase();
+            }
+
+            else if(str.length()>=3){
+                str.substring(str.length()-3).toUpperCase();
```
haithemhamid, line 12: no need for this line of code.

PR #344, opened 2025-07-13
TRA577: Back Around - Add Last Char to Front and Back of String

Shaimaalnabhani:
The backAround method adds the last letter of the string to the front and back of the string.


haithemhamid (changes requested):
else{  NewStr= str + str +str ;}
since it will be of length zero there is no value on str and it may cause error, no need for this branch of code.

PR #348, opened 2025-07-13
Front22 – Add First Two Chars to Front and Back of String

NASSER1221:
*Practiced string slicing and combining strings.


EchoCoder20 (changes requested):
You need to check if the string length is less than 2, use the whole string as the prefix and suffix otherwise, you need to get the first two Chars and add them to front and back of String.
Make sure to test different cases while running your code.
For example:
1.  System.out.println(front22("nasser"));
2.  System.out.println(front22(""));
3.  System.out.println(front22("na"));
4.  System.out.println(front22("n"));

NASSER1221:
Condition Added

PR #349, opened 2025-07-13
TRA547: EveryNth

Alfori90:
I implement a method everyNth(String str, int n) that returns a new string consisting of characters starting at index 0 and then every nth character after that. The input string is non-empty, and n is guaranteed to be 1 or more.


EchoCoder20 (approved):
Good!
Note: You do not need to use this whole code "`else if (str.isEmpty() && n >= 1) {
                return "str is empty";

            } else {
                return "n is 0";
            }` "
and instead of it you can return to the user general message on `else` block for example" `else { return "Check you input string must not be empty and n must not be negative" `

PR #351, opened 2025-07-13
StartHi – Check if String Starts with "hi"

NASSER1221:
*Practiced prefix for the strings.


EchoCoder20 (approved):
Good!
Note for next time: You need to validate that the string is not empty. Also,
Make sure to consider all possible capitalizations of the word "Hi". For example, startHi("HI nasser") should return true but with your code will return false for this input.

PR #362, opened 2025-07-13
Max1020 – Return Larger Value in Range 10..20 or 0 if None

NASSER1221:
*Practiced more ranging and conditional selectivity.


EchoCoder20 (changes requested):
Good effort, but your code is missing some conditions.
You need to check if a is within the specified range and b is not — in that case, you should return a. Otherwise, return b.

PR #363, opened 2025-07-13
StringE – Check if String Contains Between 1 and 3 'e' Characters

NASSER1221:
*Practiced counting and Strings.


EchoCoder20 (dismissed):
Well done!
In line 10, you can directly used the Char 'e' in `if` statement condition without need to declare variable.

PR #366, opened 2025-07-13
EveryNth – Build String Using Every Nth

NASSER1221:
*Practiced looping with strings.


haithemhamid (dismissed):
it will be good if you put your code to avoid any unnecessary repeated instruction as following:
```
public static String everyNth(String str , int n){
        StringBuilder String1=new StringBuilder();
        char [] w=str.toCharArray();
        for (int i=0;i<str.length();i+=n){
            String1.append(w[i]);

        }
        return String1.toString();
    }
```

NASSER1221:
Corrected

PR #370, opened 2025-07-14
TRA 682

MahaRafeet:
Adds "Not " to the beginning of the string unless it already starts with "Not" (case-insensitive).


EchoCoder20 (changes requested):
You need to check the string length if it is more than or equal 3 then you can use the substring and check if there is "not" word or not , otherwise you directly return "not"+str

MahaRafeet:
it is not required !

PR #376, opened 2025-07-14
TRA 684

MahaRafeet:
Returns a new string made by repeating the first three characters of the input string three times.


haithemhamid (changes requested):
you need to handle when the case is length of string is less than 3

MahaRafeet:
updated !

EchoCoder20 (approved):

haithemhamid (approved):

PR #383, opened 2025-07-14
TRA-855

Mulk000:
I created a Java method called startOz that checks the first two characters of a string and returns a result based on whether they are "o" and/or "z".
Extract:
First character (str.substring(0,1))
Second character (str.substring(1,2))
If the first character is "o", return it.
If the first is "o" and the second is "z", return both characters.
Otherwise, return the second character.


haithemhamid (changes requested):
You need to change the logic since if the function given this word: "ozymandias" it will return only "o" and it must return "oz"

EchoCoder20 (changes requested):
You need first to check it the string is empty then use `subString()` . Also, while using `subString` to check if it equals to "z" or "o" you need to check the length of the string

EchoCoder20 (dismissed):
Good,
Note: You can directly return "o" , "z" or "oz" without need to use `subString()`

PR #387, opened 2025-07-14
TRA 714

MahaRafeet:
Returns true if the input string starts with "hi".


haithemhamid (approved):

EchoCoder20 (approved):
Good!
Note: You do not need for code in line (8) , you can directly write the word "hi" in the if statement condition `if(firstWord.equals("hi")`

PR #395, opened 2025-07-14
TRA-925

Mulk000:
I wrote a method lastDigit that compares the last digit (units place) of two integers a and b, and returns true if they are the same, otherwise false.
I Use the modulus operator % 10 to get the last digit of each number.
If both last digits are equal, return true.
Otherwise, return false.

### DIFF
--- a/From_Mulk/BackAround.txt
+++ b/From_Mulk/BackAround.txt
@@ -1,0 +1,16 @@
+public class BackAround{
+
+      public static String backAround(String str){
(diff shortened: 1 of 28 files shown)


haithemhamid (approved):

EchoCoder20 (changes requested):
You need to be sure that a and b are not negative number

EchoCoder20 (approved):

PR #397, opened 2025-07-14
TRA 719

MahaRafeet:
Returns true if the input string has at least 3 characters and starts with "ix" at index 1.


EchoCoder20 (approved):
Note: Next time try to check if the string is empty

PR #443, opened 2025-07-27
TRA-943: warmup two

Manal-Alhatmi:


daniyalmehmood (changes requested):
This PR deletes all github files and tools. Please fix this issue.

PR #444, opened 2025-07-27
TRA-959 string one added

Hawraa-a:
String one category added


daniyalmehmood (changes requested):
Please separate the problems in separate classes and have more commits with meaningful commit messages.

PR #454, opened 2025-07-28
#tra 1062 warmup2-array667 added

sh3tha:
Count occurrences of 6 followed by 6 or 7 in array


Ghaida02a (changes requested):
there should be only 1file changed (from_shatha/Array667.java). 
no need for nested if statements, use one if statement (if (nums[i] == 6 && (nums[i + 1] == 6 || nums[i + 1] == 7)))

PR #455, opened 2025-07-28
#TRA-973: warm-up 2. StringTimes added

Ghaida02a:
Created a public method stringTimes(String str, int n) that uses a for loop and StringBuilder to concatenate the string n times.


EchoCoder20:
sync with master

from_Ghaida/stringTimes.java
```diff
@@ -1,4 +1,4 @@
-public class stringTimes {
+public class StringTimes {
```
EchoCoder20, line 1: The name of the class must be same as the name of the file

EchoCoder20 (changes requested):

PR #463, opened 2025-07-28
#tra 1060 warmup2-alt pairs added

sh3tha:
#tra 1060 warmup2-alt pairs added


Ghaida02a (changes requested):
Make the function name understandable

PR #467, opened 2025-07-28
#tra 1057warmup2-array123 added

sh3tha:
#tra 1057warmup2-array123 added


Ghaida02a (changes requested):
Function name--> it's not advised (do not add numbers)

PR #469, opened 2025-07-28
#TRA-995: warm-up 2. Last2 added

Ghaida02a:
Implemented last2(String str) method:

-If the input string has fewer than 2 characters, returns 0.

-Extracts the last 2 characters of the string.

-Loops through the string (excluding the final 2 characters) and counts how many times the same 2-character pattern appears.


Ghaida02a:
I modified the code

PR #481, opened 2025-07-28
TRA-982: stringTimes Task

kaltham11:
I created a class called JavaFundamental, then I created a function called stringDouble that takes a string and a number.
I used a StringBuilder (named str1) to store the repeated string.
Using a for loop, I appended the original string to str1 n times.
Finally, I returned the result by converting str1 to a string using .toString().


Ghaida02a (changes requested):
Nice work, you could rename str2 to something like result or doubledString—makes it slightly more semantic

Ghaida02a (approved):

PR #482, opened 2025-07-28
warmup2-TRA-981-doubleX added

sh3tha:
Check if first 'x' is immediately followed by another 'x'


Ghaida02a (changes requested):
Make the function name understandable (more descriptive).

PR #499, opened 2025-07-28
Adding String Bits

Alfori90:
Loop through the string, starting at index 0, and add every character at even indexes (0, 2, 4, ...) to a new string. Return the final result.

### DIFF
--- a/Huria/warmup2/StringBits.java
+++ b/Huria/warmup2/StringBits.java
@@ -1,0 +1,30 @@
+public class StringBits {
+
+        String str;
+
+        public static String stringBits(String str) {
+
+            String newString = "";
+
+
+                for (int i = 0; i < str.length(); i = i + 2) {
+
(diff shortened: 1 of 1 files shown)


EchoCoder20:
Sync with master!

Huria/warmup2/StringBits.java
```diff
@@ -0,0 +1,30 @@
...
+        String str;
+
+        public static String stringBits(String str) {
+
+            String newString = "";
+
+
+                for (int i = 0; i < str.length(); i = i + 2) {
+
+                    newString = newString + str.charAt(i);
```
EchoCoder20, line 12: String is immutable , you can use StringBuilder instead of string concatenation.

EchoCoder20 (changes requested):

EchoCoder20 (approved):

EchoCoder20 (approved):

PR #501, opened 2025-07-28
TRA-996: Warmup-2: Last2 added

Hawraa-a:
- I implemented the method last2 to count how many times the last 2-character substring appears elsewhere in the string (excluding the last occurrence).

- Returned 0 immediately if the string length is less than 2.

- Extracted the last 2 characters as a target substring.

- Used a for loop to check each 2-character substring except the last one, incrementing the count if it matches


Ghaida02a (approved):
Good job

PR #504, opened 2025-07-28
TRA-999: Warmup-2: ArrayCount9 added

Hawraa-a:
- I created a method arrayCount9 that counts how many times the number 9 appears in an integer array.

- Used a for loop to iterate through the array and increment a counter when an element equals 9.

- Returned the total count.


Ghaida02a (approved):
Good Job

PR #509, opened 2025-07-28
TRA-1008: Warmup-2: StringMatch added

Hawraa-a:
- I implemented the method stringMatch to count positions where two strings contain the same length-2 substring.

- Calculated the minimum length of the two strings to avoid out-of-bounds errors.

- Used a for loop to compare every 2-character substring at the same position in both strings.

- Incremented the count when substrings matched.


Ghaida02a (approved):
well done

PR #511, opened 2025-07-28
TRA-1012: Warmup-2: StringX added

Hawraa-a:
- I created the method stringX to remove all "x" characters from the string except those at the start or end.

- Used a for loop to iterate over the string characters.

- Kept characters if they are not "x" or if they are "x" at the first or last position.


Ghaida02a (approved):
well done
StringBuilder is more efficient for string concatenation in this task.

PR #514, opened 2025-07-28
TRA-1020: Warmup-2: Array667 added

Hawraa-a:
- I implemented the method array667 to count the number of times a 6 is followed by either another 6 or a 7 in an integer array.

- Used a for loop to iterate through the array up to the second last element.

- Checked if the current element is 6 and the next element is 6 or 7, then incremented the count.

- Returned the total count.


Ghaida02a (changes requested):

Ghaida02a (approved):
Well done

PR #521, opened 2025-07-28
Adding Alt Pairs

Alfori90:
I use loop through the string in steps of 4, and for each step, append the current character and the next one (if it exists) to the result string. Return the final string.


EchoCoder20:
sync with master!

EchoCoder20 (changes requested):
Please delete the files in Huria/warmup2/AltPairs.java

EchoCoder20 (approved):

PR #522, opened 2025-07-28
TRA-1126: Solving countxx task

EchoCoder20:
I create countxx() method that Count the number of "xx" in the given string.


Ghaida02a (approved):
Well done

PR #524, opened 2025-07-28
Adding Array123

Alfori90:
I use loop through the array, and at each position, check if the current and next two elements are 1, 2, and 3 in order. If found, return true; otherwise, return false.


EchoCoder20:
sync with master!

EchoCoder20 (approved):

PR #526, opened 2025-07-28
TRA-1024: Warmup-2=>The solution for arrayCount9 Task

kaltham11:
I created a class called CountNinesInArray, and a function called countNinesInArray that takes an array of integers.
The function uses a for loop to go through each element in the array.
If the element is equal to 9, it increases the count by 1.
Finally, it returns the total number of 9s found in the array.


Ghaida02a (approved):
good job

PR #530, opened 2025-07-28
frontTimes Task 

MahaRafeet:


From_Maha/frontTimes.txt
```diff
@@ -0,0 +1,13 @@
+  public String frontTimes(String str, int n) {
```
EchoCoder20, line 1: Make sure name of the function explain it's purpose

From_Maha/frontTimes.txt
```diff
@@ -0,0 +1,13 @@
...
+
+
+        String result = "";
```
EchoCoder20, line 4: String is immutable object better to use `StringBuilder` instead

EchoCoder20 (changes requested):
The file  extension must be frontTimes.java not  .txt.
Also, make sure to add class

MahaRafeet:
Updated

PR #532, opened 2025-07-28
Adding String X

Alfori90:
I use to check if the first and last characters are "x" and save them. Then remove all "x" characters from the middle part of the string. Finally, rebuild the string by combining the preserved first and last characters with the cleaned middle part.

### DIFF
--- a/Huria/warmup2/StringX.java
+++ b/Huria/warmup2/StringX.java
@@ -1,0 +1,23 @@
+public class StringX {
+    public static String stringX(String str) {
+        int len = str.length();
+        String newString = "";
+        for (int i = 0; i < len; i++) {
+            if (!(i > 0 && i < (len - 1) && str.substring(i, i + 1).equals("x"))) {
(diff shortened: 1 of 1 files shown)


Huria/warmup2/StringX.java
```diff
@@ -0,0 +1,23 @@
+public class StringX {
+    public static String stringX(String str) {
```
EchoCoder20, line 2: Consider renaming the method to a more descriptive name that clearly reflects its purpose.

Huria/warmup2/StringX.java
```diff
@@ -0,0 +1,23 @@
...
+        int len = str.length();
+        String newString = "";
```
EchoCoder20, line 4: String is immutable, try to use StringBuilder instead!

EchoCoder20 (changes requested):

EchoCoder20 (approved):

PR #535, opened 2025-07-28
TRA-1027: Warmup-2=>The solution for arrayFront9 Task

kaltham11:
I created a class called Has9InFirstFour, and a function called has9InFirstFour that takes an array of integers.
The function checks if the number 9 appears in the first four elements of the array.
It uses Math.min(nums.length, 4) to handle arrays with fewer than 4 elements.
Then it loops through the range and returns true if a 9 is found.
If not, it returns false.


Ghaida02a (approved):
Good job

PR #537, opened 2025-07-28
DoubleX Task

MahaRafeet:
This code checks if the first occurrence of `'x'` in the string is immediately followed by another `'x'`.


MahaRafeet:
Updated!

From_Maha/DoubleXChecker.java
```diff
@@ -0,0 +1,12 @@
+public class DoubleXChecker {
+
+    //Checks if the first 'x' in the string is immediately followed by another 'x'.
+
+    public boolean doubleX(String str) {
```
EchoCoder20, line 5: Make sure name of the function explain it's purpose

EchoCoder20 (changes requested):

From_Maha/DoubleXChecker.java
MahaRafeet: updated

PR #554, opened 2025-07-28
TRA-1038: The solution for altPairs Task

kaltham11:
I created a class called JavaFundamental, and a function called altPairs that takes a string.
The function builds a new string by taking characters at positions 0 and 1, skipping the next two characters, then taking 4 and 5, skipping again, and so on. It uses a for loop that increases by 4 in each step, and appends the character at i and (if it exists) the character at i + 1. Finally, it returns the result as a string.


Ghaida02a (changes requested):
Functionally correct, clean, and efficient.
*Change variable name

PR #566, opened 2025-07-28
Adding Front Times

Alfori90:
I extract the first 3 characters (or fewer if the string is shorter), then repeat that front n times to build and return the final string.

### DIFF
--- a/Huria/warmup2/FrontTimes.java
+++ b/Huria/warmup2/FrontTimes.java
@@ -1,0 +1,33 @@
+public String frontTimes(String str, int n) {
+    String front;
+    // String newString;
+    
+    if ( str.length() > 2 && n >= 0) {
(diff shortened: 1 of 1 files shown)


Huria/warmup2/FrontTimes.java
```diff
@@ -0,0 +1,33 @@
+public String frontTimes(String str, int n) {
+    String front;
+    // String newString;
+    
+    if ( str.length() > 2 && n >= 0) {
+      front = str.substring(0, 3);
+      
+      String repeatedVal = "";
```
EchoCoder20, line 8: It is better to use `StringBuilder` instead of `String`

EchoCoder20 (changes requested):
Remove unnecessary comments.

EchoCoder20 (approved):

PR #574, opened 2025-07-28
TRA-1204: Solving Array667 Task

EchoCoder20:
I created method called array667 that  pass an array of ints and return the number of times that two 6's are next to each other in the array. Also count instances where the second "6" is actually a 7.


Ghaida02a (approved):
Well done

PR #585, opened 2025-07-29
TRA-1102: String-1: helloName added

Hawraa-a:
Implemented a method helloName(String name) that takes a string and returns a greeting in the format "Hello [name]!"


Ghaida02a (approved):
Nice work!

Ghaida02a (approved):

PR #587, opened 2025-07-29
TRA-1100: String-1: makeAbba added

Hawraa-a:
Implemented makeAbba(String a, String b) to return a new string in the format "abba"


from_Nasser/name.txt
daniyalmehmood, line 1: What is the reason for you to delete "from_Nasser" folder?

daniyalmehmood (changes requested):

Ghaida02a (approved):

PR #604, opened 2025-07-29
TRA-1230-Warmup2-StringX

MahaRafeet:
This code removes all `'x'` characters from the middle of the string, keeping the first and last characters unchanged.


MahaRafeet:
Updated!

PR #620, opened 2025-07-29
TRA-1242-Warmup2-NotTriple

MahaRafeet:
This code returns `false` if the array contains any value appearing three times in a row (a "triple"); otherwise, it returns `true`.


EchoCoder20 (changes requested):
Note : 

MahaRafeet:
Updated !

PR #621, opened 2025-07-29
TRA-1243-Warmup2-Has271

MahaRafeet:
This code returns `true` if the array contains a pattern where a number is followed by that number plus 5, then a number within 2 of that number minus 1 (the "271" pattern).


EchoCoder20 (changes requested):
Note : 

MahaRafeet:
Updated !

PR #647, opened 2025-07-29
TRA-1140-Warmup2-CountXX

MahaRafeet:
This code counts how many times the substring `"xx"` appears in the input string, allowing overlaps.


From_Maha/CountXX.txt
```diff
@@ -0,0 +1,9 @@
+int countXX(String str) {
```
EchoCoder20, line 1: Make sure name of the function explain it's purpose

EchoCoder20 (changes requested):
Note : 

EchoCoder20 (dismissed):

PR #648, opened 2025-07-29
TRA-1104: String-1: extraEnd added

Hawraa-a:
- Created the extraEnd(String str) method

- Extracted the last 2 characters using substring(str.length() - 2)

- Returned a new string made of 3 copies of those 2 characters


Ghaida02a (approved):
Good job

PR #651, opened 2025-07-29
TRA-1105: String-1: firstTwo added

Hawraa-a:
- I Created the firstTwo(String str) method

- Returned the first 2 characters if the string length is 2 or more

- If the string is shorter than 2, returned it as is


Ghaida02a (approved):
Outstanding work!

PR #654, opened 2025-07-29
TRA-1108: String-1: firstHalf added

Hawraa-a:
- I Created the firstHalf(String str) method

- Calculated the midpoint using str.length() / 2

- Returned the substring from index 0 to the midpoint


Ghaida02a (approved):
Nice work

PR #672, opened 2025-07-29
TRA-1113: String-1: withouEnd2 added

Hawraa-a:
- I Created the withouEnd2(String str) method

- Removed both the first and last characters using substring(1, str.length() - 1)

- Returned an empty string if the input length is 1 or less


Ghaida02a (changes requested):

PR #700, opened 2025-07-29
TRA-1187: String Yak - Remove 'Y_k' from the String

Shaimaalnabhani:
The code returns a version of the input string with all "yak" are removed


EchoCoder20:
sync with master

PR #703, opened 2025-07-29
TRA 1397: add frontTimes

hanin190340:
The `FrontRepeater` class contains a static method `repeatFront` that takes a string and a non-negative integer `n` and returns a new string consisting of the first three characters of the input string (or the entire string if it is shorter than three characters) repeated `n` times. The method extracts the front portion of the string, then uses a loop to append it `n` times to a `StringBuilder`, finally returning the concatenated result. This function is useful for generating repeated prefixes of strings based on a given count.


EchoCoder20 (changes requested):
Note :
4. Remove unnecessary space

PR #704, opened 2025-07-29
TRA 1403: add countXX

hanin190340:
The class contains a static method that counts the number of times the substring `"xx"` appears consecutively in a given string. The method scans through the string, checking each pair of adjacent characters, and increments a counter whenever it finds `"xx"`. It then returns the total count of these overlapping occurrences. This function is useful for analyzing repeated character patterns within strings.


Ghaida02a (changes requested):
*c -->Use names that describe the role of the variable
*class name same as file name

EchoCoder20 (changes requested):
Note :
4. Remove unnecessary space

PR #707, opened 2025-07-29
TRA-1117: String-1: nTwice added

Hawraa-a:
- I Created nTwice(String str, int n) method:
- Extracts the first n characters using substring(0, n).
- Extracts the last n characters using substring(str.length() - n).
- Concatenates and returns the result.


Ghaida02a (changes requested):
(nTwice)-->make the function name understandable

Ghaida02a (approved):

PR #708, opened 2025-07-29
TRA-1118: String-1: twoChar added

Hawraa-a:
- Implemented the twoChar(String str, int index) method:

- Returns a 2-character substring starting from the given index.

- If the index is out of bounds (less than 0 or too close to the end), returns the first 2 characters of the string instead.


Ghaida02a (changes requested):
make the function name understandable

PR #709, opened 2025-07-29
TRA 1434: String splosion

hanin190340:
The DoubleXChecker class contains a static method hasDoubleX that determines whether the first occurrence of the character 'x' in a given string is immediately followed by another 'x'. The method locates the first 'x' and checks if the next character is also 'x', returning true if so, and false otherwise. This method helps detect consecutive 'x' characters at the first occurrence of 'x' within the string.


Ghaida02a (changes requested):
Class name should be the same as the file name

EchoCoder20 (changes requested):
Note :

PR #710, opened 2025-07-29
TRA 1457: stringBits

hanin190340:
The EvenIndexSelector class contains a static method selectEvenChars that returns a new string made by selecting every other character from the input string, starting with the first character (index 0). The method uses a loop that increments by 2 on each iteration, appending characters at even indices to the result. This function is useful for filtering out alternate characters to create a simplified or patterned version of the original string.


Ghaida02a (changes requested):
Good work!

from_hanin/EvenIndexSelector.txt
```diff
@@ -0,0 +1,13 @@
+public class EvenIndexSelector {
+    public static String selectEvenChars(String str) {
+        String newStr = "";
```
EchoCoder20, line 3: Convert `newstr` from `string` to `StringBuilder` .
Also, The variable name `newstr` should be renamed to something more descriptive to better reflect the data it holds.

PR #716, opened 2025-07-30
TRA-365: StringBits

Manal-Alhatmi:


From_Manal/StringBits.txt
```diff
@@ -0,0 +1,15 @@
+public class StringBits{
+public String stringBits(String str) {
+    String r ="";
```
EchoCoder20, line 3: String is immutable object better to use `StringBuilder` instead

From_Manal/StringBits.txt
```diff
@@ -0,0 +1,15 @@
+public class StringBits{
+public String stringBits(String str) {
```
EchoCoder20, line 2: Make sure name of the function explain it's purpose

PR #718, opened 2025-07-30
TRA-1121: String-1: atFirst added

Hawraa-a:
- I implemented the atFirst(String str) method inside the AtFirst class.
- I handled three conditions:
1. If the string length is 2 or more, return the first two characters using str.substring(0, 2).
2. If the string length is 1, return the character followed by '@'
3. If the string is empty, return "@@".


Ghaida02a (changes requested):
change "@" + "@"--> to "@@"

Ghaida02a (approved):

PR #719, opened 2025-07-30
TRA-1363: StringSplosion

Manal-Alhatmi:
Given a non-empty string like "Code" return a string like "CCoCodCode"


From_Manal/StringSplosion.txt
```diff
@@ -0,0 +1,15 @@
+public class StringSplosion {
+    public static String stringSplosion(String str) {
+        String r = "";
```
EchoCoder20, line 3: String is immutable object better to use `StringBuilder` instead

EchoCoder20 (changes requested):

PR #721, opened 2025-07-30
TRA-1362: StringYak

Manal-Alhatmi:
Removes Yak pattern from the string


EchoCoder20 (changes requested):

PR #723, opened 2025-07-30
TRA-1122: String-1: lastChars added

Hawraa-a:
- I implemented the lastChars(String a, String b) method.
- I extracted:
      - The first character from string a, or '@' if a is empty.
      - The last character from string b, or '@' if b is empty.
-  I returned the combined result as a new 2-character string.
- I returned the combined result as a new 2-character string.


Ghaida02a (changes requested):
*make the function name understandable (lastChars)
*you can combine if-else statements

Ghaida02a (approved):

PR #726, opened 2025-07-30
TRA-1369: FrontTimes

Manal-Alhatmi:
Given a string and a non-negative int n, we'll say that the front of the string is the first 3 chars, or whatever is there if the string is less than length 3. Return n copies of the front

### DIFF
--- a/From_Manal/ArraySumComparision.java
+++ b/From_Manal/ArraySumComparision.java
@@ -1,0 +1,18 @@
+import java.util.Arrays;
+
+public class ArraySumComparision {
+    public static int[] getBiggerOfTwoNumber(int[] a, int[] b) {
+        int sumA = a[0] + a[1];
+        int sumB = b[0] + b[1];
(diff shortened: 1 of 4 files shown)


From_Manal/FrontTimes.txt
```diff
@@ -0,0 +1,16 @@
+public class FrontTimes {
```
EchoCoder20: Make sure name of the class explain it's purpose

From_Manal/FrontTimes.txt
```diff
@@ -0,0 +1,16 @@
...
+    public static String frontTimes(String str, int n) {
+        String front = str.substring(0, Math.min(str.length(), 3));
+        String result = "";
```
EchoCoder20: Convert `result` from `string` to `stringBuilder`

EchoCoder20 (changes requested):

PR #727, opened 2025-07-30
#TRA-1280: WarmUp1_monkeyTrouble done

Hjufaili:
We have two monkeys, a and b, and the parameters aSmile and bSmile indicate if each is smiling. We are in trouble if they are both smiling or if neither of them is smiling. Return true if we are in trouble.

monkeyTrouble(true, true) → true
monkeyTrouble(false, false) → true
monkeyTrouble(true, false) → false


From_Alharith/MonkeyBehavior.java
```diff
@@ -0,0 +1,12 @@
+public class MonkeyBehavior  {
+
+    public boolean areMonkeysInTrouble (boolean aSmile, boolean bSmile) {
+        if (aSmile && bSmile){
+            return true;
+        }
+        if (!aSmile && !bSmile){
+            return true;
```
Ghaida02a, line 8: You should use else-if instead of using 2 if-statements. Or maybe you can combine them.

Ghaida02a (changes requested):

PR #728, opened 2025-07-30
TRA-1123: String-1: conCat added

Hawraa-a:
- I implemented the conCat(String a, String b) method.
- I handled different cases for concatenation:
    - If either string is empty, return the non-empty one.
    - If the last character of a is the same as the first character of b, I removed the            duplicate by appending b.substring(1) to a.
    - Otherwise, I simply concatenated both strings.


Ghaida02a (changes requested):
*Make the function name understandable
*You can combine if-statement with else if-statement if(a.isEmpty() || b.isEmpty())

Ghaida02a (approved):

PR #729, opened 2025-07-30
TRA1370: DoubleX

Manal-Alhatmi:
return true if the first instance of "x" in the string is immediately followed by another "x"


From_Manal/DoubleX.java
```diff
@@ -0,0 +1,16 @@
+public class DoubleXFinder {
```
EchoCoder20: Class name must match the filename

From_Manal/DoubleX.java
```diff
@@ -0,0 +1,16 @@
...
+    //this function check if 'X' is followed by another 'X'
+   public static boolean hasDoubleX(String str) {
+        int index= str.indexOf("x");
+        if (index== -1 || index +1>=str.length()){
```
EchoCoder20: Note that `xIndex` will never be -1 since string indices in Java always start from 0, so this check can be safely removed. Additionally, it's clearer to write the condition as `str.length() <= xIndex + 1` instead of `xIndex + 1 >= str.length()` for better readability.

EchoCoder20 (changes requested):

EchoCoder20 (dismissed):

PR #734, opened 2025-07-30
TRA 1508: Warmup-2 last2

hanin190340:
This Java class, LastTwoCharMatcher, includes a method countMatchingLastTwo that analyzes a given string and counts how many times the last two characters of that string appear earlier within the same string, excluding the final occurrence itself. For example, if the input is "hixxhi", the method looks at the last two characters "hi" and scans the rest of the string to count how many other "hi" pairs appear. This kind of logic is often used in basic text analysis or string pattern matching tasks.


Ghaida02a (changes requested):
*Avoid recreating StringBuilder unnecessarily: You're building a new StringBuilder just to call substring, but substring works directly on String objects.
*Rename variables
*Move last-two calculation outside the loop: Since it doesn’t change, calculating it once improves clarity and performance
if (str.length() < 2) return 0;
    int counter = 0;
    String lastTwo = str.substring(str.length() - 2)
before for-loop

PR #741, opened 2025-07-30
TRA-1373: ArrayCount9

Manal-Alhatmi:
Given an array of integers, this function returns the number of times the value 9 appears in the array.


EchoCoder20 (approved):
Good!

PR #745, opened 2025-07-30
TRA 1510: Warmup-2 > arrayCount9

hanin190340:
The NumberCounter class includes a method called countNines, which takes an array of integers and returns how many times the number 9 appears in that array. It loops through each element in the input array, checking for the value 9, and increments a counter each time it finds one. This method is useful for basic array analysis tasks, especially when needing to identify the frequency of a specific value.


from_hanin/NumberCounter.txt
```diff
@@ -0,0 +1,18 @@
+public class NumberCounter {
+    public static int countNines(int[] nums) {
+        int count = 0;
+
+        // pass through each number in the array
+        for (int i = 0; i < nums.length; i++) {
+            int n = nums[i];
```
EchoCoder20, line 7: The variable name `n` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):
Note :

PR #746, opened 2025-07-30
TRA-1374: Array667

Manal-Alhatmi:
Given an array of integers, this function counts the number of times the number 6 is immediately followed by either another 6 or a 7. It returns the total number of such occurrences found in the array.


From_Manal/SixSixOrSixSevenCounter.java
```diff
@@ -15,8 +15,8 @@ public static int array667(int[] nums) {
     }
```
EchoCoder20, line 15: Instead of using nested `if-statment`, you can directly merge all condition in one if statement to avoid redundancy .
`if (nums[i] == 6 && nums[i + 1] == 6)` , can be if (nums[i]==6 && nums[i+1]==6 || nums[i]==6 && nums[i+1]==7  )

EchoCoder20 (changes requested):

EchoCoder20 (dismissed):

PR #751, opened 2025-07-30
TRA-1376: AltPairs

Manal-Alhatmi:
Given a string, return a new string made by concatenating characters at indices 0 & 1, 4 & 5, 8 & 9, and so on. That is, I include two characters, skip the next two, and repeat.


From_Manal/AlternatePairsBuilder.java
```diff
@@ -0,0 +1,20 @@
+
+public class AlternatePairsBuilder {
+    public static String getAlternatesPairs(String str) {
+        //every 4 characters take index i and i+1
+        StringBuilder string1 = new StringBuilder();
```
EchoCoder20, line 5: The variable name `string1` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

EchoCoder20 (approved):

PR #760, opened 2025-07-30
TRA-1131: String-1: deFront added

Hawraa-a:
- I implemented the deFront(String str) method.
- I returned a version of the string without the first two characters unless:
    - The first character is 'a' — in which case I kept it.
    - The second character is 'b' — in which case I kept it.
- I appended the remainder of the string starting from index 2, if present.


Ghaida02a (approved):
Good job
You can use StringBuilder for more efficiency.

PR #771, opened 2025-07-30
TRA 1513:Warmup-2 (array123)

hanin190340:
The SequenceChecker class includes the method contains123Sequence, which scans an array of integers to determine whether the sequence 1, 2, 3 appears in order and consecutively. It loops through the array and checks each triplet of consecutive values. If it finds 1, 2, and 3 in that order, it returns true; otherwise, it returns false. This function is helpful when looking for specific patterns or sequences in integer arrays.


EchoCoder20 (changes requested):
Note :

PR #775, opened 2025-07-30
TRA 1514: Warmup-2 > string Match

hanin190340:
The String Match class provides a method called countMatchingSubstrings that compares two input strings and counts how many positions contain identical two-character substrings. It iterates through both strings up to the length of the shorter one (minus one), extracting a substring of length 2 from each at every position. If the substrings match at the same position in both strings, it increments the count. This is helpful for detecting aligned patterns or partial matches between strings, such as in text comparison or similarity scoring.


Ghaida02a (changes requested):
*Format the code
*Rename variables (str, str1, str2)

EchoCoder20 (changes requested):
Good work!
Note :

PR #778, opened 2025-07-30
No Triples - Return false  if the array contains a triple.

Shaimaalnabhani:
The code returns true if the given array does not contain any triples.


EchoCoder20:
sync with master!

PR #799, opened 2025-07-30
TRA 1517: Warmup-2 > altPairs

hanin190340:
The Selective Pair class defines a method selectEveryOtherPair that builds a new string from an input by taking characters at specific intervals. It includes the characters at indexes 0 and 1, then skips the next two, includes 4 and 5, skips again, and so on (i.e., 0–1, 4–5, 8–9, ...). This method is useful when you need to sample short segments of a string in a repeating pattern, such as when parsing encoded data or creating stylized output.


from_hanin/altPairs.txt
```diff
@@ -0,0 +1,15 @@
+public class SelectivePair {
+    public static String selectEveryOtherPair(String str) {
+        StringBuilder str1=new StringBuilder();
```
EchoCoder20, line 3: The variable name `str1` should be renamed to something more descriptive to better reflect the data it holds.

from_hanin/altPairs.txt
```diff
@@ -0,0 +1,15 @@
+public class SelectivePair {
```
EchoCoder20, line 1: Class name must match the filename

EchoCoder20 (changes requested):
Note :
2. Format your code

PR #814, opened 2025-07-30
TRA 1518: Warmup-2 > stringYak

hanin190340:
The YakRemover class provides a method called stringYak that removes all occurrences of the pattern "y_k" from a given string, where the first character is 'y', the third is 'k', and the character in between can be anything. It loops through the string and skips over any such pattern when found. This approach is useful for cleaning or filtering out specific patterned substrings—like "unlucky" or flagged sequences—while preserving the rest of the content intact.


from_hanin/stringYak.txt
```diff
@@ -0,0 +1,15 @@
+public class YakRemover {
```
EchoCoder20, line 1: Class name must match the filename

from_hanin/stringYak.txt
```diff
@@ -0,0 +1,15 @@
...
+    public static String stringYak(String str) {
```
EchoCoder20, line 2: Make sure name of the function and class explain it's purpose

from_hanin/stringYak.txt
```diff
@@ -0,0 +1,15 @@
...
+        StringBuilder str1 = new StringBuilder(str);
```
EchoCoder20, line 3: The variable name `str1` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):
Note :

PR #827, opened 2025-07-30
TRA 1520: Warmup-2 > noTriples

hanin190340:
The TripleChecker class provides a method called hasNoTriples that determines whether an input array contains any "triples"—that is, three consecutive elements with the same value. The method iterates through the array and checks every group of three adjacent numbers. If such a group is found, it immediately returns false; otherwise, it returns true. This function is useful for validating sequences that should not contain repeated patterns or errors.


from_hanin/noTriples.txt
```diff
@@ -0,0 +1,11 @@
+public class TripleChecker {
```
EchoCoder20, line 1: Class name must match the filename

EchoCoder20 (changes requested):
Note :

PR #831, opened 2025-07-30
TRA-1527: stringBits

haithemhamid:
Given a string, return a new string made of every other char starting with the first, so "Hello" yields "Hlo".


from_Haitham/StringBits.java
```diff
@@ -1,9 +1,9 @@
-public class stringBits {
+public class StringBits {
 public String stringBits(String str) {
         String newString = "";
         for (int i = 0; i < str.length(); i++) {
             if (i % 2 == 0) newString += str.charAt(i);
```
EchoCoder20, line 5: Instead of using `if-statment` inside loop you can directly increased the value of `i` by 2 in the for loop (i+=2) , which is more efficient.

EchoCoder20 (changes requested):

PR #861, opened 2025-07-30
TRA1198: String Splosion - Create a string from growing prefixes

Shaimaalnabhani:
The code takes a non-empty string and returns a new string made by concatenating substrings of increasing length starting from the first character.

### DIFF
--- a/From_Shaima/StringGrowth.java
+++ b/From_Shaima/StringGrowth.java
@@ -1,0 +1,16 @@
+public class StringGrowth {
+    public static String generateStringGrowth(String str)
+    {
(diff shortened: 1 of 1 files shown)


From_Shaima/StringGrowth.java
```diff
@@ -0,0 +1,16 @@
+public class StringGrowth {
+    public static String generateStringGrowth(String str)
+    {
+        StringBuilder newStr= new StringBuilder(); // Create a StringBuilder object to build 'newStr' variable
```
EchoCoder20, line 4: The variable name `newStr` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

EchoCoder20 (approved):

PR #904, opened 2025-07-30
TRA1487: Warmup-2 > countXX

NASSER1221:
I'm given a string to count the number of times the substring "xx" appears in the string.
also showing that "xx" substrings can overlap.


from_Nasser/CountXX.java
```diff
@@ -0,0 +1,17 @@
+public class CountXX {
+
+    public static int countXX(String str) {
```
EchoCoder20, line 3: Make sure name of the function explain it's purpose

EchoCoder20 (changes requested):

from_Nasser/CountXX.java
NASSER1221: function name changed

PR #940, opened 2025-07-30
TRA 1644: String-1 > make Abba

hanin190340:
The Abba Formatter class includes a method build Abba String that takes two strings and returns a new string arranged in the pattern "abba" — meaning it concatenates the first string, then the second string twice, and then the first string again. This format is useful in playful message construction, string formatting, or simple text pattern exercises.


Ghaida02a (changes requested):
1- Filename should be same as class name.
2- File extension must be .java not .txt.

PR #945, opened 2025-07-30
TRA-1547: has271

haithemhamid:
Given an array of ints, return true if it contains a 2, 7, 1 pattern: a value, followed by the value plus 5, followed by the value minus 1. Additionally the 271 counts even if the "1" differs by 2 or less from the correct value.


from_Haitham/Has271.java
```diff
@@ -1,4 +1,4 @@
-public class has271 {
+public class Has271 {
```
EchoCoder20, line 1: Make sure name of the Class explain it's purpose. Also, Class should not contain numbers

from_Haitham/Has271.java
```diff
@@ -1,4 +1,4 @@
...
 public boolean has271(int[] nums) {
```
EchoCoder20, line 2: Make sure name of the Class explain it's purpose

EchoCoder20 (changes requested):

PR #966, opened 2025-07-30
TRA 1645:String-1 > makeTags

hanin190340:
The Html Tag Builder class includes a method makeTags that dynamically constructs an HTML element by wrapping a given word with the specified tag. It takes the tag name (e.g., "b", "i", or "cite") and a string of content, then returns the HTML-formatted string with the appropriate opening and closing tags. This is useful for generating simple HTML elements in templating systems, web development, or text formatting tools.


Ghaida02a (changes requested):
1- Filename should be same as class name.
2- File extension must be .java not .txt.

PR #1010, opened 2025-07-30
#TRA-1332: String-2. xyzMiddle added

Ghaida02a:
Implemented xyzMiddle(String str) method:

-Checks if the string contains the substring "xyz" positioned near the middle.
-Iterates through the string up to str.length() - 3 to safely extract substrings of length 3.
-For each occurrence of "xyz", calculates the number of characters to the left and right.
-Returns true if the counts on both sides differ by at most 1, ensuring "xyz" is near the center.
-Returns false otherwise or if the string length is less than 3.

### DIFF
--- a/from_Ghaida/xyzMiddle.java
+++ b/from_Ghaida/xyzMiddle.java
@@ -1,0 +1,23 @@
+public class xyzMiddle {
+    public static void main(String[] args) {
+        System.out.println(xyzMiddle("AAxyzBB")); //true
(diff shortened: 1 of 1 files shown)


from_Ghaida/xyzMiddle.java
```diff
@@ -0,0 +1,23 @@
+public class xyzMiddle {
```
daniyalmehmood, line 1: You are violating Java class naming conventions

daniyalmehmood (changes requested):

PR #1011, opened 2025-07-30
Warmup-1: #TRA-1626 in1020 done

mohammedanad-dev:
Given 2 int values, return true if either of them is in the range 10..20 inclusive.

in1020(12, 99) → true
in1020(21, 12) → true
in1020(8, 99) → false

### DIFF
--- a/from_mohammed/in1020.java
+++ b/from_mohammed/in1020.java
@@ -1,0 +1,7 @@
+public class in1020 {
+    public boolean in1020(int a, int b) {
+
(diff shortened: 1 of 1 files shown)


from_mohammed/in1020.java
```diff
@@ -0,0 +1,7 @@
+public class in1020 {
```
daniyalmehmood, line 1: You are violating Java class naming conventions

daniyalmehmood (changes requested):

PR #1019, opened 2025-07-30
Warmup-1: #TRA-1628 loneTeen done

mohammedanad-dev:
We'll say that a number is "teen" if it is in the range 13..19 inclusive. Given 2 int values, return true if one or the other is teen, but not both.

loneTeen(13, 99) → true
loneTeen(21, 19) → true
loneTeen(13, 13) → false

### DIFF
--- a/from_mohammed/loneTeen.java
+++ b/from_mohammed/loneTeen.java
@@ -1,0 +1,6 @@
+public class loneTeen {
+    public boolean loneTeen(int a, int b) {
+        boolean aa=(a>=13&&a<=19);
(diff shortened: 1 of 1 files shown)


from_mohammed/loneTeen.java
```diff
@@ -0,0 +1,6 @@
+public class loneTeen {
```
daniyalmehmood, line 1: Violating java class naming conventions

daniyalmehmood (changes requested):

PR #1023, opened 2025-07-30
#TRA-1336: String-2. zipZap added

Ghaida02a:
Implemented zipZap(String str):

-Iterates through the string character by character.
-Checks if the current character is 'z' and the character two positions ahead is 'p'.
-If so, appends 'z' and 'p' to the new string and skips the middle character.
-Otherwise, appends the current character as-is.
-Handles edge cases to avoid IndexOutOfBoundsException.


daniyalmehmood:
Violates class naming convention

PR #1095, opened 2025-07-30
TRA 1657: String-1 > withouEnd2

hanin190340:
The String Cutter class contains the without End method, which returns a version of a string with the first and last characters removed. It ensures the string is at least two characters long, throwing an exception if it's too short or null. This method is useful when you need to strip boundary characters (like quotes, brackets, or padding) from a string during processing or formatting.


from_hanin/WithoutEnd 2.txt
```diff
@@ -0,0 +1,10 @@
+public class StringCutter {
+    public String withoutEnd(String str) {
+        if(str.length()>=2)
+        {
```
daniyalmehmood, line 4: Format your code

daniyalmehmood (changes requested):

PR #1102, opened 2025-07-30
TRA1254: The solution for MakeOutWord Task

kaltham11:
I created a class called MakeOutWord, and a function called makeOutWord that takes two strings:
out: a 4-character string (like "<<>>" or "**##")
word: the word to be inserted
The function returns a new string by placing the word between the first two and last two characters of out.
It uses substring(0, 2) for the beginning, and substring(2, 4) for the ending.


Ghaida02a (approved):
Good job

PR #1180, opened 2025-07-31
TRA-1774: Solving String-1 First Half Task

EchoCoder20:
I created getFirstHalfOfString method that pass a string of even length and return the first half. So the string "WooHoo" yields "Woo".


Ghaida02a (approved):
Well done

PR #1236, opened 2025-07-31
TRA1219: Front Times - Repeat the first three characters n times

Shaimaalnabhani:
The code returns a string made by repeating the front part of the input string n times. The “front” is defined as the first 3 characters of the string, or the entire string if its length is less than 3

### DIFF
--- a/From_Shaima/Repeater.java
+++ b/From_Shaima/Repeater.java
@@ -1,0 +1,34 @@
+public class Repeater  {
+    public static String repeatFront(String str, int n) {
+        StringBuilder front = new StringBuilder(); // Create a StringBuilder object to build 'front' variable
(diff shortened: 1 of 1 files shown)


From_Shaima/Repeater.java
```diff
@@ -0,0 +1,34 @@
+public class Repeater  {
+    public static String repeatFront(String str, int n) {
+        StringBuilder front = new StringBuilder(); // Create a StringBuilder object to build 'front' variable
```
EchoCoder20, line 3: The variable name `front` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

PR #1252, opened 2025-07-31
TRA1179: Array667 - Count how many times the array contains 66 or 67

Shaimaalnabhani:
The code returns the number of times the array contains 6 followed by 6 or followed by 7.

### DIFF
--- a/From_Shaima/ArrayPatternChecker667.java
+++ b/From_Shaima/ArrayPatternChecker667.java
@@ -1,0 +1,18 @@
+public class ArrayPatternChecker667 {
+    public static int count66And67(int[] nums) {
+        int count =0; // initialize  'count' variable and assign 0 to its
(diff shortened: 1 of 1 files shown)


From_Shaima/ArrayPatternChecker667.java
```diff
@@ -0,0 +1,18 @@
+public class ArrayPatternChecker667 {
```
EchoCoder20, line 1: Class should not have numbers

From_Shaima/ArrayPatternChecker667.java
```diff
@@ -0,0 +1,18 @@
...
+    public static int count66And67(int[] nums) {
```
EchoCoder20, line 2: Make sure name of the function explain it's purpose

EchoCoder20 (changes requested):

EchoCoder20 (approved):

PR #1295, opened 2025-07-31
Right2 file

MahaRafeet:
Added right2 method that moves the last two characters of a string to the beginning


MahaRafeet:
Updated!

PR #1361, opened 2025-08-03
TRA1897: First Half - Return the  first half of the string

Shaimaalnabhani:
The code returns a string of the first half of the string.


EchoCoder20:
Sync with master

PR #1363, opened 2025-08-03
TRA1536: Adding Without First Characters Task

Alfori90:
The code takes two input strings and returns their concatenation, excluding the first character of each string. The input strings are guaranteed to be at least 1 character long, making this operation safe.


from_Huria/SkipFirstChars.java
```diff
@@ -0,0 +1,17 @@
+public class SkipFirstChars {
+    public static String skipFirstChars(String a, String b) {
+        String newString = a.substring(1, a.length()) + b.substring(1, b.length());
```
EchoCoder20, line 3: The variable name `newString` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

PR #1366, opened 2025-08-03
TRA1515: Adding Left Rotate by 2Characters Task

Alfori90:
The code takes a string and returns a new version where the first 2 characters are moved to the end of the string. The input string is guaranteed to be at least 2 characters long.

### DIFF
--- a/from_Huria/RotateLeft2.java
+++ b/from_Huria/RotateLeft2.java
@@ -1,0 +1,21 @@
+public class RotateLeft2 {
+    public static String rotateLeft2(String str) {
+        if (str.length() <= 2) {
+            return str;
+        } else {
+            String first = str.substring(0, 2);
+            String second = str.substring(2, str.length());
+            String newString = second + first;
+            return newString;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


from_Huria/RotateLeft2.java
```diff
@@ -0,0 +1,21 @@
+public class RotateLeft2 {
+    public static String rotateLeft2(String str) {
```
EchoCoder20, line 2: The class and function name must not contain numbers

from_Huria/RotateLeft2.java
```diff
@@ -0,0 +1,21 @@
...
+        if (str.length() <= 2) {
+            return str;
+        } else {
+            String first = str.substring(0, 2);
+            String second = str.substring(2, str.length());
+            String newString = second + first;
```
EchoCoder20, line 8: The variable name `newString` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

PR #1368, opened 2025-08-03
TRA1534: Adding Middle Two Character Extractor Task

Alfori90:
The code takes a string of even length and returns a new string consisting of its middle two characters.

### DIFF
--- a/from_Huria/Mid2Picker.java
+++ b/from_Huria/Mid2Picker.java
@@ -1,0 +1,22 @@
+public class Mid2Picker {
+    public static String mid2Picker(String str) {
+        if (str.length() <= 1) {
+            return "";
+        }
+        int len = str.length() / 2;
+        String middTwo = str.substring(len - 1, len + 1);
+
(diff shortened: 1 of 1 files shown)


from_Huria/Mid2Picker.java
```diff
@@ -0,0 +1,22 @@
+public class Mid2Picker {
+    public static String mid2Picker(String str) {
```
EchoCoder20, line 2: The class and function name must not contain numbers

from_Huria/Mid2Picker.java
```diff
@@ -0,0 +1,22 @@
...
+        if (str.length() <= 1) {
+            return "";
+        }
+        int len = str.length() / 2;
```
EchoCoder20, line 6: The variable name `len` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

PR #1372, opened 2025-08-03
solution of StringTimes task

Mulk000:
In this code, I created a class called StringTimes with a method named stringTimes that takes a string and an integer n. The method uses a StringBuilder to repeat the input string n times by appending it in a loop. After the loop finishes, it returns the final combined string using toString().


Ghaida02a (dismissed):
Good job

PR #1376, opened 2025-08-03
solution of FrontTimes task

Mulk000:
In this code, I created a class called FrontTimes with a method named createRepeatedText that takes a string and an integer n. The method checks if the string is empty and returns an empty string if it is. Otherwise, it repeats the first three characters of the string (or the whole string if it's shorter than three characters) n times using a loop and appends the result to a StringBuilder, which is then returned as the final string.


Ghaida02a (approved):
Well done

PR #1380, opened 2025-08-03
TRA-1263: String-1=>The Solution for RightTwo Task

kaltham11:
I created a class called RightTwo, and a function called right2 that takes a string str.
The function checks if the string length is 2 or less; if so, it returns the string unchanged.
Otherwise, it moves the last two characters to the front by taking a substring of the last two characters and appending the rest of the string (excluding the last two).
This effectively rotates the last two characters to the front of the string.


Ghaida02a (changes requested):
(right2) function name--> it's not advised (do not add numbers)

Ghaida02a (approved):

PR #1382, opened 2025-08-03
TRA-1265: String-1=>The Solution for withouEnd2 Task

kaltham11:
I created a class called WithouEndTwo, and a function called withouEnd2 that takes a string str.
The function checks if the string has a length of 3 or more.
If so, it returns the string without the first and last characters using substring(1, str.length() - 1).
If the string is shorter than 3 characters, it returns an empty string.


Ghaida02a (changes requested):
Make the function and class name understandable (more descriptive).
function name--> it's not advised (do not add numbers)

Ghaida02a (approved):

PR #1394, opened 2025-08-03
TRA-1274: The Solution for LastChars Task

kaltham11:
I created a class called LastChars, and a function called lastChars that takes two strings a and b.
The function returns a new two-character string made of the first character of a and the last character of b.
If a is empty, it uses "@" as the first character.
If b is empty, it uses "@" as the second character.
If both strings are empty, it returns "@@".
This ensures the result always contains exactly two characters.

### DIFF
--- a/from_Kaltham/LastChars.java
+++ b/from_Kaltham/LastChars.java
@@ -1,0 +1,15 @@
+public class LastChars {
+    public String lastChars(String a, String b) {
+        if (a.isEmpty() && b.isEmpty()) {
+            return "@" + "@";
+        }
+        if (a.isEmpty()) {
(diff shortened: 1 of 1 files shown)


from_Kaltham/LastChars.java
```diff
@@ -0,0 +1,15 @@
+public class LastChars {
+    public String lastChars(String a, String b) {
+        if (a.isEmpty() && b.isEmpty()) {
+            return "@" + "@";
+        }
```
Ghaida02a, line 5: you can write it like this--> return "@@"

Ghaida02a (changes requested):
Make the function name and class name understandable (more descriptive).

PR #1395, opened 2025-08-03
TRA-1275: String-1=>The Solution for ConCat Task

kaltham11:
I created a class called ConCat, and a function called conCat that takes two strings a and b.
The function concatenates the two strings with one exception:
If the last character of a is the same as the first character of b, it removes the duplicate character from the end of a before concatenating.
If either string is empty, it returns the non-empty one.
If both are empty, it returns an empty string.


Ghaida02a (changes requested):
Make the function name understandable (more descriptive).

Ghaida02a (approved):
Well  done

PR #1399, opened 2025-08-03
TRA-1278: The Solution for FrontAgain Task

kaltham11:
I created a class called FrontAgain, and a function called frontAgain that takes a string str.
The function checks if the first two characters of the string are the same as the last two characters.
It uses substring(0, 2) and substring(str.length() - 2) for comparison.
If they match, it returns true; otherwise, it returns false.


Ghaida02a (changes requested):
Make the function name understandable

Ghaida02a (approved):
Good job

PR #1412, opened 2025-08-03
solution of  StringSplosion task

Mulk000:
In this code, I created a class called StringSplosion with a method named explodeString that takes a string as input. The method builds a new string by repeatedly adding substrings that start at the beginning of the original string and increase in length by one character each time. This way, it creates a “splosion” effect by combining all these substrings into one final string, which it then returns.


Ghaida02a (approved):
Well done

PR #1415, opened 2025-08-03
solution of Last2 task

Mulk000:
In this code, I created a class called Last2 with a method named countRepeatingLast2 that takes a string as input. The method looks at the last two characters of the string and counts how many times that same two-character sequence appears elsewhere in the string (excluding the very end). It does this by looping through the string and comparing each two-character substring to the last two characters, incrementing a counter when they match. Finally, it returns the total count.


Ghaida02a (changes requested):
*Class violates naming conventions
*countX--> the variable name is not understandable

PR #1421, opened 2025-08-03
solution of ArrayCountNine task

Mulk000:
In this code, I created a class called `ArrayCountNine` with a method named `count9InArray` that takes an array of integers as input. The method loops through the array and counts how many times the number 9 appears. It increments a counter each time it finds a 9, then returns the total count at the end.


Ghaida02a (changes requested):
*countX--> make it understandable
*Function name --> it's not advised (do not use numbers).

PR #1422, opened 2025-08-03
solution of ArrayFrontNine task

Mulk000:
In this code, I created a class called `ArrayFrontNine` with a method named `firstFourContains9` that takes an array of integers as input. The method checks the first four elements (or fewer if the array is shorter) to see if any of them is the number 9. If it finds a 9 within that range, it returns `true`; otherwise, it returns `false`.


Ghaida02a (changes requested):
*Function name--> it's not advised (do not add numbers).

PR #1424, opened 2025-08-03
solution of  ArrayOneTwoThree task 

Mulk000:
In this code, I created a class called `ArrayOneTwoThree` with a method named `contains123Sequence` that takes an array of integers as input. The method checks through the array to see if the sequence `1, 2, 3` appears consecutively anywhere. If it finds this sequence, it returns `true`; otherwise, it returns `false`.


Ghaida02a (changes requested):
*Function name --> it's not advised (do not add numbers)
*ArrayOneTwoThree --> make it understanable class name

PR #1425, opened 2025-08-03
solution of  StringMatch task

Mulk000:
In this code, I created a class called `StringMatch` with a method named `countMatchingPairs` that takes two strings as input. The method compares both strings by checking pairs of two characters at the same positions in each string. It counts how many times these pairs match and returns the total count. To avoid errors, it only checks up to the length of the shorter string minus one.


Ghaida02a (changes requested):
The variables name like (mid, countX) area bit unclear

PR #1434, opened 2025-08-03
solution of AltPairs task

Mulk000:
In this code, I created a class called `AltPairs` with a method named `getAlternatePairs` that takes a string as input. The method builds a new string by repeatedly taking two characters at a time, starting at every 4th character in the original string. It appends the current character and, if available, the next character to a `StringBuilder`. Finally, it returns the combined result as the new string.


Ghaida02a (dismissed):
Good job

PR #1462, opened 2025-08-04
Warmup2: #TRA-2048 StringTimes done

mohammedanad-dev:
Given a string and a non-negative int n, return a larger string that is n copies of the original string.

stringTimes("Hi", 2) → "HiHi"
stringTimes("Hi", 3) → "HiHiHi"
stringTimes("Hi", 1) → "Hi"

### DIFF
--- a/from_Mohammed/StringTimes.java
+++ b/from_Mohammed/StringTimes.java
@@ -1,0 +1,6 @@
+public String StringTimes(String str, int n) {
+    String res="";
+    for(int i=0;i<n;i++)
+        res+=str;
+    return res;
+}


from_Mohammed/StringTimes.java
```diff
@@ -0,0 +1,6 @@
+public String StringTimes(String str, int n) {
+    String res="";
```
EchoCoder20, line 2: Convert `res` from String to StringBuilder.
The variable name 'res' should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):
Note :

PR #1464, opened 2025-08-04
Warmup2: #TRA-2049 FrontTimes done

mohammedanad-dev:
Given a string and a non-negative int n, we'll say that the front of the string is the first 3 chars, or whatever is there if the string is less than length 3. Return n copies of the front;

frontTimes("Chocolate", 2) → "ChoCho"
frontTimes("Chocolate", 3) → "ChoChoCho"
frontTimes("Abc", 3) → "AbcAbcAbc"

### DIFF
--- a/from_Mohammed/FrontTimes.java
+++ b/from_Mohammed/FrontTimes.java
@@ -1,0 +1,10 @@
+public String FrontTimes(String str, int n) {
+
+    int frontl =3;
+    if (frontl>str.length())
+        frontl=str.length();
+    String front = str.substring(0, frontl);
+    String res = "";
+    for(int i=0;i<n;i++)
+        res +=front;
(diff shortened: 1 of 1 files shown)


EchoCoder20 (commented):
Note :
3. Make sure name of the function and Class explain it's purpose

from_Mohammed/FrontTimes.java
```diff
@@ -1,4 +1,5 @@
-public String FrontTimes(String str, int n) {
+public class FrontTimes{
+public String frontTimes(String str, int n) {
```
EchoCoder20, line 2: Make sure name of the function and Class explain it's purpose

EchoCoder20 (changes requested):

PR #1491, opened 2025-08-04
Warmup-2: #TRA-2056 arrauCount9 done

mohammedanad-dev:
Given an array of ints, return the number of 9's in the array.

arrayCount9([1, 2, 9]) → 1
arrayCount9([1, 9, 9]) → 2
arrayCount9([1, 9, 9, 3, 9]) → 3

### DIFF
--- a/from_Mohammed/ArrayCount9.java
+++ b/from_Mohammed/ArrayCount9.java
@@ -1,0 +1,8 @@
+public int ArrayCount9(int[] nums) {
+    int count=0;
+    for (int i=0; i<nums.length; i++)
+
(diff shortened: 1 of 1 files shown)


from_Mohammed/ArrayCount9.java
```diff
@@ -0,0 +1,8 @@
+public int ArrayCount9(int[] nums) {
```
EchoCoder20, line 1: The name of the function must follow camleCase format. Also, Make sure name of the function explain it's purpose

EchoCoder20 (changes requested):
Note :

PR #1496, opened 2025-08-04
Warmup-2: #TRA-2057 arrayFront9 done

mohammedanad-dev:
Given an array of ints, return true if one of the first 4 elements in the array is a 9. The array length may be less than 4.

arrayFront9([1, 2, 9, 3, 4]) → true
arrayFront9([1, 2, 3, 4, 9]) → false
arrayFront9([1, 2, 3, 4, 5]) → false

### DIFF
--- a/from_Mohammed/ArrayFront9.java
+++ b/from_Mohammed/ArrayFront9.java
@@ -1,0 +1,10 @@
+public boolean ArrayFront9(int[] nums) {{
+    int done=nums.length;
+    if (done>4) done=4;
+
+    for (int i=0; i<done; i++)
+
+        if (nums[i]==9)
+
+            return true;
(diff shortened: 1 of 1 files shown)


from_Mohammed/ArrayFront9.java
```diff
@@ -0,0 +1,10 @@
+public boolean ArrayFront9(int[] nums) {{
+    int done=nums.length;
+    if (done>4) done=4;
+
+    for (int i=0; i<done; i++)
```
EchoCoder20, line 5: Consider using `nums.length` directly rather than storing it in a separate variable.

EchoCoder20 (changes requested):
Note :

PR #1498, opened 2025-08-04
Warmup-2: #TRA-2058 array123 done

mohammedanad-dev:
Given an array of ints, return true if the sequence of numbers 1, 2, 3 appears in the array somewhere.

array123([1, 1, 2, 3, 1]) → true
array123([1, 1, 2, 4, 1]) → false
array123([1, 1, 2, 1, 2, 3]) → true

### DIFF
--- a/from_Mohammed/Array123.java
+++ b/from_Mohammed/Array123.java
@@ -1,0 +1,8 @@
+public boolean Array123(int[] nums) {
+    {
+        for (int i=0; i<(nums.length-2);i++)
+
+            if (nums[i]==1&&nums[i+1]==2&&nums[i+2]==3)
+                return true;
+
+    }return false;}


from_Mohammed/Array123.java
```diff
@@ -1,8 +1,10 @@
-public boolean Array123(int[] nums) {
+public class Array123{
```
EchoCoder20, line 1: Make sure name of the Class explain it's purpose. Also, class name must not contain numbers

from_Mohammed/Array123.java
```diff
@@ -1,8 +1,10 @@
...
+public boolean array123(int[] nums) {
```
EchoCoder20, line 2: Make sure name of the function explain it's purpose. Also, function name is better not contain numbers

EchoCoder20 (changes requested):

PR #1500, opened 2025-08-04
Warmup-2: #TRA-2059 StringMatch done

mohammedanad-dev:
Given 2 strings, a and b, return the number of the positions where they contain the same length 2 substring. So "xxcaazz" and "xxbaaz" yields 3, since the "xx", "aa", and "az" substrings appear in the same place in both strings.

stringMatch("xxcaazz", "xxbaaz") → 3
stringMatch("abc", "abc") → 2
stringMatch("abc", "axc") → 0

### DIFF
--- a/from_Mohammed/StringMatch.java
+++ b/from_Mohammed/StringMatch.java
@@ -1,0 +1,12 @@
+
+public int StringMatch(String a, String b) {
+    int leng= Math.min (a.length(),b.length());
+    int tot=0;
+    for (int i=0; i<leng-1; i++){
+        String aSub=a.substring(i,i+2);
+        String bSub=b.substring(i,i+2);
+        if (aSub.equals(bSub))
+            tot++;
+    }
+    return tot;
+}


from_Mohammed/StringMatch.java
```diff
@@ -1,5 +1,5 @@
-
-public int StringMatch(String a, String b) {
+public class StringMatch{
```
EchoCoder20, line 1: Make sure name of the Class explain it's purpose

from_Mohammed/StringMatch.java
```diff
@@ -1,5 +1,5 @@
-
-public int StringMatch(String a, String b) {
+public class StringMatch{ 
+public int stringMatch(String a, String b) {
```
EchoCoder20, line 2: Make sure name of the function explain it's purpose

from_Mohammed/StringMatch.java
```diff
@@ -1,5 +1,5 @@
...
     int leng= Math.min (a.length(),b.length());
```
EchoCoder20, line 3: The variable name `leng` should be renamed to something more descriptive to better reflect the data it holds.

from_Mohammed/StringMatch.java
```diff
@@ -1,5 +1,5 @@
...
     int tot=0;
```
EchoCoder20, line 4: The variable name `tot` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

PR #1505, opened 2025-08-04
Warmup-2: #TRA-2062 AltPairs done

mohammedanad-dev:
Given a string, return a string made of the chars at indexes 0,1, 4,5, 8,9 ... so "kittens" yields "kien".

altPairs("kitten") → "kien"
altPairs("Chocolate") → "Chole"
altPairs("CodingHorror") → "Congrr"

### DIFF
--- a/from_Mohammed/AltPairs.java
+++ b/from_Mohammed/AltPairs.java
@@ -1,0 +1,9 @@
+public String AltPairs(String str) {
+    String res="";
+    for (int i=0; i<str.length();i+=4){
+        int end=i+2;
+        if (end>str.length()){
+            end=str.length();}
+        res+=str.substring(i,end);}
+    return res;
+}


from_Mohammed/AltPairs.java
```diff
@@ -1,9 +1,10 @@
-public String AltPairs(String str) {
+public class AltPairs{
```
EchoCoder20, line 1: Make sure name of the Class explain it's purpose

from_Mohammed/AltPairs.java
```diff
@@ -1,9 +1,10 @@
...
+public String altPairs(String str) {
```
EchoCoder20, line 2: Make sure name of the function explain it's purpose

from_Mohammed/AltPairs.java
```diff
@@ -1,9 +1,10 @@
...
     String res="";
```
EchoCoder20, line 3: The variable name `res` should be renamed to something more descriptive to better reflect the data it holds.
Also, convert String to StringBuilder

from_Mohammed/AltPairs.java
```diff
@@ -1,9 +1,10 @@
...
     for (int i=0; i<str.length();i+=4){
         int end=i+2;
```
EchoCoder20, line 5: The variable name `end` should be renamed to something more descriptive to better reflect the data it holds.

EchoCoder20 (changes requested):

PR #1512, opened 2025-08-04
Warmup-2: #TRA-2065 NoTriples\ done

mohammedanad-dev:
Given an array of ints, we'll say that a triple is a value appearing 3 times in a row in the array. Return true if the array does not contain any triples.

noTriples([1, 1, 2, 2, 1]) → true
noTriples([1, 1, 2, 2, 2, 1]) → false
noTriples([1, 1, 1, 2, 2, 2, 1]) → false

### DIFF
--- a/from_Mohammed/NoTriples.java
+++ b/from_Mohammed/NoTriples.java
@@ -1,0 +1,6 @@
+public boolean NoTriples(int[] nums){ {
+    for(int i=0;i<(nums.length-2);i++){
+        int first=nums[i];
+        if(nums[i+1]==first&&nums[i+2]==first) return false;
+    }
+}return true;}


from_Mohammed/NoTriples.java
```diff
@@ -1,6 +1,8 @@
-public boolean NoTriples(int[] nums){ {
+public class NoTriples{
```
EchoCoder20, line 1: Make sure name of the class explain it's purpose

from_Mohammed/NoTriples.java
```diff
@@ -1,6 +1,8 @@
...
+public boolean noTriples(int[] nums){ {
```
EchoCoder20, line 2: Make sure name of the function explain it's purpose

from_Mohammed/NoTriples.java
```diff
@@ -1,6 +1,8 @@
...
     for(int i=0;i<(nums.length-2);i++){
```
EchoCoder20, line 3: You can remove unnecessary brackets that is around` nums.length-1 `

from_Mohammed/NoTriples.java
```diff
@@ -1,6 +1,8 @@
...
         int first=nums[i];
```
EchoCoder20, line 4: You can directly used `nums[i]`  instead to assign its value in `first` variable

EchoCoder20 (changes requested):

PR #1532, opened 2025-08-04
solution of ExtraEnd task

Mulk000:
I built a method that validates string length and replicates its last two characters three times to form a new string. It returns a message if the input is shorter than two characters.

### DIFF
--- a/From_Mulk/ExtraEnd.java
+++ b/From_Mulk/ExtraEnd.java
@@ -1,0 +1,9 @@
+public class ExtraEnd {
+    public String replicateStringEnding(String str) {
+        if (str.length() >= 2) {
+            return str.substring(str.length() - 2) + str.substring(str.length() - 2) + str.substring(str.length() - 2);
+        } else {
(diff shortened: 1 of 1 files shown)


From_Mulk/ExtraEnd.java
```diff
@@ -0,0 +1,9 @@
+public class ExtraEnd {
+    public String replicateStringEnding(String str) {
+        if (str.length() >= 2) {
+            return str.substring(str.length() - 2) + str.substring(str.length() - 2) + str.substring(str.length() - 2);
+        } else {
+            return "length less than 2";
+        }
```
Ghaida02a, line 7: It should return str. 
you can write it without else { }
if (str.length() >= 2) {
            return str.substring(str.length() - 2) + str.substring(str.length() - 2) + str.substring(str.length() - 2);
        }
            return "length less than 2"; }

Ghaida02a (changes requested):

PR #1968, opened 2025-08-10
#TRA-2323: Logic-1. shareDigit added

Ghaida02a:
This PR introduces the checkShareDigits(int a, int b) method to determine whether two two-digit integers share at least one digit. The method:

-Extracts the tens and units digits of both numbers.
-Compares all combinations to check for any shared digit.
-Returns true if at least one digit matches; otherwise, returns false.

### DIFF
--- a/from_Ghaida/ShareDigitChecker.java
+++ b/from_Ghaida/ShareDigitChecker.java
@@ -1,0 +1,19 @@
+public class ShareDigitChecker {
+    public static void main(String[] args) {
+        System.out.println(checkShareDigits(12, 23));// → true
+        System.out.println(checkShareDigits(12, 43));// → false
+        System.out.println(checkShareDigits(12, 44));// → false
+    }
+    public static boolean checkShareDigits(int a, int b) {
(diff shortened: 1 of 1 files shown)


from_Ghaida/ShareDigitChecker.java
```diff
@@ -0,0 +1,19 @@
...
+        System.out.println(checkShareDigits(12, 23));// → true
+        System.out.println(checkShareDigits(12, 43));// → false
+        System.out.println(checkShareDigits(12, 44));// → false
+    }
+    public static boolean checkShareDigits(int a, int b) {
+        int leftDigitOfA = a/10;
+        int leftDigitOfB = b/10;
+        int rightDigitOfA = a%10;
+        int rightDigitOfB = b%10;
+        if(a >= 10 && a <= 99 && b >= 10 && b <= 99) {
```
haithemhamid, line 12: No need to make this line because requirement of question already give you that the input will be in range of 10 to 99.

haithemhamid (changes requested):
The question gives you that

PR #3718, opened 2025-08-27
TRA2577: Warmup-1 > sumDouble

JamalZeidi:
Create a method that takes two integer values and returns their sum. However, if the two values are equal, return double their sum instead. This means the method should check for equality and adjust the result accordingly.

Examples:

sumDouble(1, 2) → 3 (normal sum)

sumDouble(3, 2) → 5 (normal sum)

sumDouble(2, 2) → 8 (double of 2 + 2)


EchoCoder20 (changes requested):
1.  **File:** From_Jamal/SumOfValues.java
    **Line:** 4
    **Comment:** The opening brace for the method body should typically be on the same line as the method signature or on the next line, but the spacing here is inconsistent with standard Java conventions. More importantly, there is excessive vertical spacing before the conditional check.
    **Example:**
    
```java
    public static int sumDouble(int a, int b) {
        if (a == b){
    // ...
    ```

---
2.  **File:** From_Jamal/SumOfValues.java
    **Line:** 6
    **Comment:** The opening brace for the `if` block is on a separate line and has an unnecessary extra newline after it. Standard Java practice (often called the K&R style or a variation) places the opening brace on the same line as the `if` statement, and the closing brace should align with the `if`.
    **Example:**
    
```java
    if (a == b) {
        return (a + b) * 2;
    } else {
        return (a + b);
    }
    ```

---
3.  **File:** From_Jamal/SumOfValues.java
    **Line:** 8-10
    **Comment:** There are excessive blank lines within the logic block, which reduces readability. The return statement should immediately follow the opening brace or be cleanly separated.
    **Example:**
    
```java
    if (a == b){
        return (a+b)*2;
    }
    ```

---
4.  **File:** From_Jamal/SumOfValues.java
    **Line:** 12
    **Comment:** The closing brace for the `if` block is followed by an unnecessary blank line before the `else` block starts. Also, the braces for the `if-else` structure are inconsistent with standard formatting (see point 2).
    **Example:**
    
```java
    } else {
        return (a + b);
    }
    ```

---
5.  **File:** From_Jamal/SumOfValues.java
    **Line:** 16
    **Comment:** The `main` method content has excessive vertical spacing between the two `println` calls.
---
6.  **File:** From_Jamal/SumOfValues.java
    **Line:** 1-20
    **Comment:** The class declaration should be followed by a standard single newline before the first method, not an empty line. Also, consider adding Javadoc to the `sumDouble` method to explain what it does, its parameters, and what it returns, especially since the logic involves a specific doubling condition.

    **Example (Method Signature with Javadoc):**
    
```java
    /**
     * Calculates the sum of two integers. If the integers are equal, the sum is doubled.
     *
     * @param a the first integer
     * @param b the second integer
     * @return the sum of a and b, or double the sum if a equals b
     */
    public static int sumDouble(int a, int b) {
    // ...
    ```

---
7.  **File:** From_Jamal/SumOfValues.java
    **Line:** 3-14
    **Comment:** The logic inside `sumDouble` can be simplified by removing the explicit `if-else` structure, especially since the `else` block is just returning the basic sum. You can calculate the sum first and then apply the multiplier.

    **Example (Simplified Logic):**
    
```java
    public static int sumDouble(int a, int b) {
        int sum = a + b;
        if (a == b) {
            sum *= 2;
        }
        return sum;
    }
    ```

PR #3720, opened 2025-08-27
TRA2579: Warmup-1 > diff21

JamalZeidi:
Create a method that takes an integer n and returns the absolute difference between n and 21. However, if n is greater than 21, return double the absolute difference instead. This means the method must check whether n is over 21 and adjust the result accordingly.

Examples:

diff21(19) → 2 (normal difference)

diff21(10) → 11 (normal difference)

diff21(21) → 0 (no difference)

diff21(25) → 8 (double of 25 − 21 = 4)


EchoCoder20 (changes requested):
1. **File:** From_Jamal/DifferenceOfNumbers.java
**Line:** 5
**Comment:** The method name `diff21` is not very descriptive. Consider a name that clearly indicates what the method calculates, such as `calculateDifferenceFrom21`.

---
2. **File:** From_Jamal/DifferenceOfNumbers.java
**Line:** 7-15
**Comment:** The formatting with excessive newlines makes the code harder to read. Consolidate conditional logic onto fewer lines where appropriate, adhering to standard Java style guides (like Oracle's). Also, `Math.abs(n-21)` is redundant when `n <= 21` because `n - 21` will always be zero or negative, so negating it (or using `21 - n`) is sufficient.

**Example:**

```java
    public static int calculateDifferenceFrom21(int n) {
        if (n <= 21) {
            return 21 - n;
        } else {
            return (n - 21) * 2;
        }
    }
```

---
3. **File:** From_Jamal/DifferenceOfNumbers.java
**Line:** 10
**Comment:** When `n` is exactly 21, `n - 21` is 0, and `Math.abs(0)` is 0. If `n > 21`, the requirement seems to be doubling the difference, but the logic `return Math.abs(n-21)*2;` is correct for the difference *if* the logic intended is "if $n>21$, return $2 \times (n-21)$". However, based on common coding challenges this function often implements: if $n \le 21$, return $21-n$; otherwise, return $2 \times (n-21)$. Since you are using `Math.abs` in both branches, let's simplify the first branch as discussed above, and ensure the second branch correctly implements the required doubling. If $n > 21$, $n-21$ is positive, so `Math.abs(n-21)` is just $n-21$.

---
4. **File:** From_Jamal/DifferenceOfNumbers.java
**Line:** 18-20
**Comment:** Standard practice is to only have one statement per line, and avoid empty lines between method calls in `main` unless separating logical groups of tests.

**Example:**

```java
        System.out.println(diff21(21));
        System.out.println(diff21(11));
        System.out.println(diff21(30)); // Added a test case for n > 21
    }
```

PR #3736, opened 2025-08-27
TRA2614: Warmup-1 > close10

JamalZeidi:
Create a method that takes two integer values and returns the one closest to 10. Use Math.abs(n - 10) to calculate the distance from 10. If both values are equally close, return 0 to indicate a tie.

Examples:

close10(8, 13) → 8 (distance: 2 vs. 3)

close10(13, 8) → 8 (same as above, order reversed)

close10(13, 7) → 0 (both are 3 away from 10)


EchoCoder20 (changes requested):
**File:** From_Jamal/CloseToTen.java
**Line:** 5
**Comment:** Java standard practice for class names is PascalCase (e.g., `CloseToTen`). The current naming is correct for a class name. However, the file name should match the class name exactly (case-sensitive in some file systems). Assuming the file name is `CloseToTen.java`, this is fine.

---
**File:** From_Jamal/CloseToTen.java
**Line:** 7
**Comment:** Variables names like *aAbs* and *bAbs* could be more descriptive, perhaps indicating what they represent in relation to the number 10.

---
**File:** From_Jamal/CloseToTen.java
**Line:** 7, 8
**Comment:** Excessive vertical whitespace makes the code harder to read. Remove unnecessary blank lines.
**Example:**

```java
        int aAbs = Math.abs(a - 10);
        int bAbs = Math.abs(b - 10);
```


---
**File:** From_Jamal/CloseToTen.java
**Line:** 10-14
**Comment:** The closing brace for the first if-block is on a new line, which is unconventional for a single-line block. It's better to keep the braces on the same line if the block is short, or follow a consistent style. Also, remove the unnecessary blank lines inside the block.
**Example:**

```java
        if (aAbs < bAbs) {
            return a;
        }
```


---
**File:** From_Jamal/CloseToTen.java
**Line:** 15-19
**Comment:** Similar to the previous point, remove unnecessary vertical whitespace inside the else if block.
**Example:**

```java
        else if (bAbs < aAbs) {
            return b;
        }
```


---
**File:** From_Jamal/CloseToTen.java
**Line:** 21
**Comment:** The logic returns 0 if *aAbs* equals *bAbs* (i.e., if both numbers are equally close to 10). The problem statement for "CloseToTen" usually implies returning 0 only if both are equally close. This implementation correctly returns 0 in that case, but the formatting around the final return statement can be cleaned up.

---
**File:** From_Jamal/CloseToTen.java
**Line:** 20-24
**Comment:** Remove unnecessary blank lines surrounding the final return statement.
**Example:**

```java
        return 0;
```


---
**File:** From_Jamal/CloseToTen.java
**Line:** 1
**Comment:** Java classes should generally have spacing conventions applied. Add a newline at the end of the file for better compatibility with some tools.

---
**File:** From_Jamal/CloseToTen.java
**Line:** 27
**Comment:** In the `main` method, the test case `close10(10, 100)` should ideally be accompanied by comments or assertions explaining the expected outcome for better testability and clarity. Since 10 is distance 0 from 10, and 100 is distance 90 from 10, the expected output is 10.

PR #3738, opened 2025-08-27
TRA2613 : Warmup-1 > intMax

JamalZeidi:
Create a method that takes three integer values—a, b, and c—and returns the largest of the three. Use conditional comparisons or built-in functions to determine which value is greatest.

Examples:

intMax(1, 2, 3) → 3

intMax(1, 3, 2) → 3

intMax(3, 2, 1) → 3


EchoCoder20 (changes requested):
**File:** From_Jamal/MaxNumber.java
**Line:** 5
**Comment:** The class declaration should be on the same line as the `public class MaxNumber` declaration if there are no Javadoc comments, or if Javadoc comments are present, they should precede the class declaration directly. Also, the empty lines after the class declaration are excessive.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 5 - 29
**Comment:** There are numerous excessive blank lines throughout the file, which negatively impact readability. Please remove unnecessary whitespace.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 8 - 10
**Comment:** The variables `maxNum` and `maxNum2` are calculated but never used for the final result in this method. They can be removed. The variable `maxNum3` is also unused and can be removed.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 8
**Comment:** Consider using a more descriptive name for the variable holding the maximum value, although since it's immediately overwritten or replaced by `max`, it's less critical here, but generally, avoid calculating intermediate results that are not fully utilized.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 10
**Comment:** The variable `maxNum2` is an unnecessary intermediate calculation. Also, the calculation `Math.max(b,c)` is redundant since `max` already calculates `Math.max(b,c)` internally.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 11
**Comment:** The variable `maxNum3` is unused and can be removed.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 13
**Comment:** The entire logic can be simplified to just this single line, making the method much cleaner.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 16 - 25
**Comment:** There are excessive blank lines before the return statement, which should be cleaned up for better readability.

---
**File:** From_Jamal/MaxNumber.java
**Line:** 13
**Comment:** The entire logic inside `intMax` can be reduced to a single statement for finding the maximum of three integers, making it more concise.

**Example:**

```java
    public static int intMax(int a, int b, int c) {
        return Math.max(a, Math.max(b, c));
    }
```

PR #3751, opened 2025-08-27
TRA2586: Warmup-1 > frontBack

JamalZeidi:
Create a method that takes a string and returns a new string where the first and last characters are swapped. If the string has length 1 or less, return it unchanged.

Examples:

frontBack("code") → "eodc"

frontBack("a") → "a"

frontBack("ab") → "ba"


EchoCoder20 (changes requested):
**File:** `From_Jamal/ExchangeLetters.java`
**Line:** (Whole file)
**Comment:** The code contains multiple excessive blank lines within the class and method definitions. This affects readability by making the code unnecessarily sparse. Please remove the extra blank lines for a cleaner and more compact presentation.
**Example:**

```java
public class ExchangeLetters {

    public static String frontBack(String str) {
        if (str.length() <= 1) {
            return str;
        }
        String first = str.substring(0, 1);
        String middle = str.substring(1, str.length() - 1);
        String last = str.substring(str.length() - 1);
        return last + middle + first;
    }

        System.out.println(frontBack("Jamal"));
    }
}
```

---
**File:** `From_Jamal/ExchangeLetters.java`
**Line:** `if(str.length() <= 1 ) return str;{ ... }`
**Comment:** This line has an unusual and confusing structure. The semicolon after `return str` terminates the `if` statement's action. The following empty braces `{}` are a separate, standalone code block that does nothing. This should be rewritten as a standard `if` statement for clarity and correctness.
**Example:**

```java
if (str.length() <= 1) {
    return str;
}
```

---
**File:** `From_Jamal/ExchangeLetters.java`
**Line:** `// The word will be inverted,end of the code !! ,`
**Comment:** The comment in the `main` method is unnecessary and slightly inaccurate. The method does not invert the word, but rather swaps the first and last letters. Comments like "end of the code" are considered noise and should be removed to keep the code clean.

PR #3770, opened 2025-08-27
TRA2592: Warmup-1 > or35

JamalZeidi:
Create a method that takes a non-negative integer and returns true if the number is a multiple of 3 or 5. Use the modulo operator % to check divisibility:

num % 3 == 0 checks if it's divisible by 3

num % 5 == 0 checks if it's divisible by 5

Examples:

or35(3) → true (divisible by 3)

or35(10) → true (divisible by 5)

or35(8) → false (not divisible by 3 or 5)


EchoCoder20 (changes requested):
**File:** `From_Jamal/FirstTwoChar.java`
**Line:** `public static String front22(String str)`
**Comment:** The method's readability can be improved by removing the excessive blank lines. The code can also be made more concise.
---
**File:** `From_Jamal/FirstTwoChar.java`
**Line:**

```java
int num =2 ;




if (num > str.length()){

    num = str.length();


}

String order = str.substring(0,num);
```

**Comment:** The logic to handle strings shorter than 2 characters can be simplified using `Math.min`. Additionally, the variable names `num` and `order` are not very descriptive. Using names like `prefixLength` and `prefix` would make the code's purpose clearer. There is also a minor formatting issue with an extra space before the semicolon in `int num =2 ;`.
**Example:**

```java
public static String front22(String str) {
    int prefixLength = Math.min(2, str.length());
    String prefix = str.substring(0, prefixLength);
    return prefix + str + prefix;
}
```

PR #3774, opened 2025-08-27
TRA2597: Warmup-1 > icyHot

JamalZeidi:
Create a method that takes two temperature values (integers) and returns true if one is less than 0 and the other is greater than 100. This means you're checking for an extreme cold and extreme hot condition simultaneously.

Examples:

icyHot(120, -1) → true (hot and cold)

icyHot(-1, 120) → true (cold and hot)

icyHot(2, 120) → false (only hot, no cold)


EchoCoder20 (changes requested):
**File:** `From_Jamal/CompareTwoTemp.java`
**Line:**

```java
if ((temp1 > 100 || temp2 > 100) && (temp1 < 0 || temp2 < 0) ){

    return true;

}else return false;
```

**Comment:** The method contains several unnecessary blank lines, which affect readability. Additionally, an `if-else` statement that returns `true` or `false` based on a condition is redundant. You can return the boolean expression directly to make the code more concise and clear. The logical condition can also be written in a way that is easier to understand at a glance.
**Example:**

```java
public static boolean icyHot(int temp1, int temp2) {
    return (temp1 < 0 && temp2 > 100) || (temp1 > 100 && temp2 < 0);
}
```

PR #4072, opened 2025-08-28
Functional-2 > noNeg task

Mulk000:
This method removes all negative numbers from a list of integers and returns the updated list.

### DIFF
--- a/From_Mulk/NoNegative.java
+++ b/From_Mulk/NoNegative.java
@@ -1,0 +1,8 @@
+//Functional-2 > noNeg
+
+public class NoNegative {
+    public List<Integer> removeNegativeNumbers(List<Integer> nums) {
+        nums.removeIf(n -> n < 0);
+        return nums;
+    }
+}


Ghaida02a (commented):
**File:** From_Mulk/NoNegative.java
**Line:** 4
**Comment:** The method `removeNegativeNumbers` modifies its input parameter, `nums`, directly using `nums.removeIf()`. According to the specified rule, function parameters passed to a method must not be modified inside that method. This can lead to unexpected side effects for the caller of this method. Instead, consider creating and returning a new list that contains only the non-negative numbers, leaving the original list untouched.
**Example:**
```java
import java.util.ArrayList;
import java.util.List;
import java.util.stream.Collectors;

public class NoNegative {
    public List<Integer> filterNonNegativeNumbers(List<Integer> numbers) {
        // Option 1: Using streams for a functional approach
        return numbers.stream()
                      .filter(n -> n >= 0)
                      .collect(Collectors.toList());

        // Option 2: Creating a copy and then removing (less efficient than streams for this case)
        // List<Integer> result = new ArrayList<>(numbers);
        // result.removeIf(n -> n < 0);
        // return result;
    }
}
```

PR #4075, opened 2025-08-28
Add array2 sameEnds

MahaRafeet:
This method verifies if the first len elements of an array match the last len elements.
It returns true only if both ends are identical.

### DIFF
--- a/From_Maha/ArrayEndsChecker.java
+++ b/From_Maha/ArrayEndsChecker.java
@@ -1,0 +1,10 @@
+public class ArrayEndsChecker {
+    public boolean hasSameEnds(int[] nums, int len) {
+        for (int i = 0; i < len; i++) {
+            if (nums[i] != nums[nums.length - len + i]) {
+                return false;
+            }
+        }
+        return true;
+    }
+}


EchoCoder20 (changes requested):
**File:** `From_Maha/ArrayEndsChecker.java`
**Line:** `public boolean hasSameEnds(int[] nums, int len)`
**Comment:** The method lacks validation for the `len` parameter. If `len` is negative or greater than the length of the `nums` array, the code will throw an `ArrayIndexOutOfBoundsException`. You should add a check to handle these edge cases gracefully.
**Example:**

```java
public boolean hasSameEnds(int[] nums, int len) {
    if (nums == null || len < 0 || len > nums.length) {
        return false; // Or throw an IllegalArgumentException
    }
    for (int i = 0; i < len; i++) {
        if (nums[i] != nums[nums.length - len + i]) {
            return false;
        }
    }
    return true;
}
```

---
**File:** `From_Maha/ArrayEndsChecker.java`
**Line:** `public boolean hasSameEnds(int[] nums, int len)`
**Comment:** The parameter name `len` is a bit generic. A more descriptive name like `length` or `prefixLength` would make the method's purpose clearer at a glance, improving the code's readability.
**Example:**

```java
public boolean hasSameEnds(int[] nums, int length) {
    // ...
}
```

PR #4296, opened 2025-09-02
Adding canBalance file

MahaRafeet:
Implemented isBalanced method which returns true if there is a place to split the array into two parts with equal sums.
The method calculates the total sum first, then iterates to check if the sum of the left part equals the sum of the right part.

### DIFF
--- a/From_Maha/ArrayBalanceChecker.java
+++ b/From_Maha/ArrayBalanceChecker.java
@@ -1,0 +1,17 @@
+public class ArrayBalanceChecker {
+    public boolean isBalanced(int[] numbers) {
+        int totalSum = 0;
+        for (int i = 0; i < numbers.length; i++) {
+            totalSum += numbers[i];
+        }
+        int sumLeft = 0;
+        for (int i = 0; i < numbers.length; i++) {
+            sumLeft += numbers[i];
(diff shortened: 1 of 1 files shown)


EchoCoder20 (changes requested):
**File:** `From_Maha/ArrayBalanceChecker.java`
**Line:** `for (int i = 0; i < numbers.length; i++) {`
**Comment:** The loop for calculating the sum can be made more concise and readable by using an enhanced for-loop (for-each loop), as the index `i` is not needed for the summing operation.
**Example:**

```java
int totalSum = 0;
for (int number : numbers) {
    totalSum += number;
}
```


---
**File:** `From_Maha/ArrayBalanceChecker.java`
**Line:** `for (int i = 0; i < numbers.length; i++) {`
**Comment:** The loop condition `i < numbers.length` allows an iteration where the "split" happens after the last element, making the right side of the array empty. Typically, a balanced split implies two non-empty parts. To enforce this, the loop should only run up to the second-to-last element.
**Example:**

```java
for (int i = 0; i < numbers.length - 1; i++) {
    sumLeft += numbers[i];
    int sumRight = totalSum - sumLeft;
    if (sumLeft == sumRight) {
        return true;
    }
}
```

PR #4463, opened 2025-09-08
Adding eordLen file

MahaRafeet:
Creates a map from an array of strings where each string is a key 
and its value is the length of that string. Returns the map.


Ghaida02a (changes requested):
Please provide the Java code you would like me to review. I'm ready to help

